PR #237, opened 2024-05-06
Added Authentication

### LINKED ISSUE #110: Authentication
- [ ] Simple login page
- [ ] Save token locally 
- [ ] Integrate token into requests

Corepex:
Resolves #110 
Related to https://github.com/pimcore/studio-backend-bundle/pull/91


vin0401:
I have to admit that the folder structure confuses me a bit and makes it hard to see how the things are connected.
My suggestion would be the following:

```
app/
  router/
    router.ts // <- takes care of the route handling. Maybe also offers the possibility to register routes early on before the are processed by the createBrowserRouter. 
   
components
   // if any presentational or purely UI components are available place them here.

modules/
  app/
    default-page.ts // new suffix -page to indicate a react component that is registered via module as a page.
  auth/
    // everything related to the auth module. Mostly the parts under componets/login-form
    // and also the auth-slice should go here. That way we can group all files based on the feature.
```

vin0401:
Any updates due to the folder structure? 
For me it still would make sense to group everything by feature in a new module called "auth": https://github.com/pimcore/studio-ui-bundle/pull/237#issuecomment-2122329091

vin0401 (approved):

PR #324, opened 2024-05-29
Added `Properties` view

### LINKED ISSUE #182: Properties
https://www.figma.com/file/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?type=design&node-id=490-59345&mode=design&t=l4k909oyeYybtcsS-4
https://www.figma.com/file/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?type=design&node-id=2123-117952&mode=design&t=l4k909oyeYybtcsS-4

image

Corepex:
Resolves #182


assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
```diff
@@ -0,0 +1,77 @@
...
+        defaultValue={ props.getValue() }
+        onBlur={ onBlur }
+        onChange={ saveValue }
+        onKeyDown={ onKeyDown }
+        open={ open }
+        options={ [
+          { value: 'jack', label: 'Jack' },
+          { value: 'lucy', label: 'Lucy' },
+          { value: 'Yiminghe', label: 'yiminghe' }
+        ] }
```
vin0401: Is it right that this is still hardcoded?
What are we missing here?

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
```diff
@@ -0,0 +1,77 @@
...
+          { props.getValue() }
+        </>
+      )
+    }
+
+    function onBlur (e: React.FocusEvent<HTMLInputElement>): void {
+      saveValue(e.target.value)
+    }
+
+    function onKeyDown (e: React.KeyboardEvent<HTMLInputElement>): void {
```
vin0401: Would be nice when we also could add the enter key here to save the value.
Like we do it for the text field.

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/boolean-cell/boolean-cell.tsx
```diff
@@ -0,0 +1,71 @@
...
+    function onKeyDown (e: React.KeyboardEvent<HTMLInputElement>): void {
+      if (e.key === 'Escape') {
+        disableEditMode()
+      }
+    }
+
+    return (
+      <Checkbox
+        checked={ props.getValue() }
+        onChange={ saveValue }
```
vin0401: I would only send the update for the data when hitting enter.
Currently the value would directly send via update event even so when I try to abort with escape.

So maybe we use the original cell value just as an initial value for the component state. 
Hitting enter then sends the state value per update event.

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
```diff
@@ -0,0 +1,77 @@
+/**
```
vin0401: Maybe we can move it to the grid folder?
For now this looks quite reusable for different types of grids. :)

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/boolean-cell/boolean-cell.tsx
```diff
@@ -0,0 +1,71 @@
+/**
```
vin0401: Maybe we can move this file to the grid folder. For now this looks quite reusable. :)

vin0401 (changes requested):
Looks good already. I just have some small improvements I would suggest, here. :)

markus-moser:
@Corepex Things I noticed during testing:

- [x]  Adding custom properties is not yet implemented. The properties can be saved later together with the asset, so the add button should just add the new row to the table/form:

image

- [x]  Adding predefined properties not implemented yet (same like custom properties):
image

- [x]  As the "All Properties" tab is a ready-only view I would remove the "predefined properties" select and "Add a custom property" button there and show it only in the "edit own properties" tab.
image

- [x] The properties tab is currently only visible for image assets. It should be visible for folders, text assets etc. too

- [x] When I close and re-open an asset the properties information get's not reloaded from the server. We need to refresh the data on each (re-)load.

- [x] When I edit the value of for example a text field the value get's not updated in the table (not sure if related to this PR or a "problem" of the inline edit functionality itself)

- [x] The open button currently is not implemented yet:

image

- [x] The delete button is missing in the "Actions" column:
image

Should it really be an "Actions" column if we display the "open" button directly in the "value" column and therefore only the "Delete" action is present. The column should be called "Delete", right?

- [x] Inline edit of (inheritable) checkboxes currently not implemented:

- [x] The "All Properties" tab is a readonly view. Therefore the inline edit functionallity needs to be disabled and the "Actions" column should be hidden.
- [x] (Nice2Have) When the properties list is empty the "All properties" view looks a little bit strange, do we have an idea how to improve it?

image

- [x] We need to find a way for our table grid to handle long content:
image

It does not make sense to make the tables that narrow when we have a lot of content. Don't know what the best approaches are, but we should use the width of the panel if the content requests it.

- [x] Should the checkbox value column really be centered? For me this looks a bit strange in combination with other types:
image

- [x] The "Description" column must not be editable

- [x] The "Description" column should display the word "Custom" for custom properties

- [x] "Select" property options are hardcoded:
image

- [x] "Select type for properties has no icon:

image

Please let me know if I should split this up into multiple issues...

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
Corepex: I was thinking about the translation of the stings. 
We only get a comma-separated list from the BE, something like "foo,bar,bazz." 
Wasn't sure how I should handle that - I implemented a basic display of the predefined options for now. (see https://github.com/pimcore/studio-ui-bundle/pull/324/commits/24cc73433484e68ddd6653de91fcacacabf06e2d)

@markus-moser, does Pimcore have an option to translate the options of a select in the predefined props?

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
Corepex: Good point. Implemented with https://github.com/pimcore/studio-ui-bundle/pull/324/commits/0f835c21c2a1154f35528d784d7020b665836a1f

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/boolean-cell/boolean-cell.tsx
Corepex: Not sure about the usability here. My idea was to send the changes each time you check the checkbox (because obv. it is a change). Don`t know if the "send on enter" conflicts with the keyboard navigation.

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
Corepex: done with https://github.com/pimcore/studio-ui-bundle/pull/324/commits/458d218286356d64030e6c25af5c0d3fe91beeaf

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/boolean-cell/boolean-cell.tsx
Corepex: done with https://github.com/pimcore/studio-ui-bundle/pull/324/commits/458d218286356d64030e6c25af5c0d3fe91beeaf

Corepex:
@markus-moser according your changes:

**Adding custom properties is not yet implemented. The properties can be saved later together with the asset, so the add button should just add the new row to the table/form.**

TBI


**Adding predefined properties not implemented yet (same like custom properties).**

TBI


**As the "All Properties" tab is a ready-only view I would remove the "predefined properties" select and "Add a custom property" button there and show it only in the "edit own properties" tab.**

removed the mentioned buttons on "all" tab. (see https://github.com/pimcore/studio-ui-bundle/pull/324/commits/5e071cbb62a44035169f4e8180a63dc6cd92cbb9)


**The properties tab is currently only visible for image assets. It should be visible for folders, text assets etc. too**

TBI



**When I close and re-open an asset the properties information get's not reloaded from the server. We need to refresh the data on each (re-)load.**

TBI


**When I edit the value of for example a text field the value get's not updated in the table (not sure if related to this PR or a "problem" of the inline edit functionality itself)**

TBI


**The open button currently is not implemented yet:**

True, that's because we can potentially have links to `documents` or `dataobjects` and this will break the application no - should we only implement it for assets (for now)?


**The delete button is missing in the "Actions" column**

changed icon in https://github.com/pimcore/studio-ui-bundle/pull/324/commits/16984ea8f7a78a25f97494a62011b3dc01da9708 according to #326 


**Inline edit of (inheritable) checkboxes currently not implemented**

TBI


**The "All Properties" tab is a readonly view. Therefore the inline edit functionallity needs to be disabled and the "Actions" column should be hidden.**

TBI


**(Nice2Have) When the properties list is empty the "All properties" view looks a little bit strange, do we have an idea how to improve it?**

The table will only be shown if there at least one entry. (see https://github.com/pimcore/studio-ui-bundle/pull/324/commits/a3247e83897b83d1a03d6a53fab235ff811aaad4)


**We need to find a way for our table grid to handle long content:**

I already talked to @vin0401 about the grid scaling issue. There is also no option for now to make the grid full-width.
https://github.com/pimcore/studio-ui-bundle/pull/324/commits/e7a90477e4b7ab34dee512c7b5787111a4484909


**Should the checkbox value column really be centered? For me this looks a bit strange in combination with other types:**

IMHO It also looks weird when its only on the left ... don't know what the best solution is ...


**The "Description" column must not be editable**

see https://github.com/pimcore/studio-ui-bundle/pull/324/commits/5f4a4b028b1962736b31900016148e0f68cc135a


**The "Description" column should display the word "Custom" for custom properties**

Added the `name` column. (https://github.com/pimcore/studio-ui-bundle/pull/324/commits/d8c9417332ef60247c29de7cbd55dd01624bff7f, https://github.com/pimcore/studio-ui-bundle/pull/324/commits/36c8920700a05d47ba46c4a490e1931668baed71)


**"Select" property options are hardcoded:**

please see: https://github.com/pimcore/studio-ui-bundle/pull/324#discussion_r1621948480


**"Select type for properties has no icon:**

That is true; this icon has to be redesigned by @idaiv. - added with https://github.com/pimcore/studio-ui-bundle/pull/324/commits/8ce9702228d903ce453c54796e7474f2bfd4a524 and https://github.com/pimcore/studio-ui-bundle/pull/324/commits/4425112e0eb939135427a34c81d45f1d7192604d

https://github.com/pimcore/studio-ui-bundle/pull/326

assets/js/src/core/modules/asset/editor/image/tab-manager/tabs/properties/components/table/cells/select-cell/select-cell.tsx
markus-moser: @Corepex In the classic UI it's not translatable. So let's leave it as not translateable for now.

Corepex: Thanks for checking @markus-moser. I guess than the current implementation should do it :)

vin0401:
@Corepex - In regards to this point:

> We need to find a way for our table grid to handle long content:
>
>I already talked to @vin0401 about the grid scaling issue. There is also no option for now to make the grid full-width.

We can't make it full width but at least we can add some additional handling for long content.
And we already have that for the asset listing. 

image

I guess there is a new wrapper container that breaks the css in that case. But a few adaptions done to this container should already fix that. Reach out to me when you need help. :)

idaiv:
to @markus-moser points:
- Should it really be an "Actions" column if we display the "open" button directly in the "value" column and therefore only the "Delete" action is present. The column should be called "Delete", right? ---> yes, when there is a single action like Delete, then the column label is based on it. If there are multiple actions, then the label is "Actions"
- (Nice2Have) When the properties list is empty the "All properties" view looks a little bit strange, do we have an idea how to improve it? ---> We can add a message for all empty tables in general "No context/data yet." to inform the user. And still keep the column headings for a reference what content is supposed to be in such a table. 
- "Select type for properties has no icon ---> yes, we have discussed with Christian and couldnt find an example on the current UI of such. I should create a new one without a reference then? 

And Markus, please let me know if I missed anything from the list or you need certain examples.

markus-moser:
@idaiv Thanks for your input 👍 

The select icon looks like this in the current UI:

image

Maybe you have an idea for the checkbox topic?

image

markus-moser:
@Corepex @vin0401 I added follow-up tickets for the remaining properties topics - let's continue our work based on those issues:

#329 
#330 
#331 
#332 
#333 
#334

PR #325, opened 2024-05-31
[Feature] Added `embedded metadata` tab

### LINKED ISSUE #151: Embedded Metainfo
see https://github.com/pimcore/product-management/issues/376

image

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=3021-127407&t=roQfjNiUiWGvKejZ-0

Allow detaching

Corepex:
Resolves #151 
Blocked by #237


Corepex:
There is an issue with one of the endpoints. I will double-check that with @lukmzig.

Corepex:
> There is an issue with one of the endpoints. I will double-check that with @lukmzig.

fixed with https://github.com/pimcore/studio-backend-bundle/pull/82

markus-moser:
@Corepex Not sure why but I have a problem with the authentication for the API when I try it with this branch together with the latest 1.x of the backend:

image

Is there something which we can check or should we wait until the session auth is done (should be done relatively soon on backend side)?

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/embedded-metadata/embedded-metadata-container.styles.tsx
```diff
@@ -0,0 +1,27 @@
...
+*  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import { createStyles } from 'antd-style'
+
+export const useStyles = createStyles(({ css, token }) => {
+  return {
+    tab: css`
+        p:first-of-type {
```
vin0401: This rule seems to be a bit to weak for me. 

When we get any kind of paragraph in the grid or maybe any other additions under this tab that involves a paragraph, the styles will automatically applied on the first one.

Corepex: Which other parameter should be in that tab and above the headline?
But yeah i will wrap it in a div :)

vin0401: Currently it applies to any first <p> in every element in the tab. 
``` html
<div class="i-am-a-grid-cell">
  <p></p>
</div>
```
Also this would be affected by the styles. I just want to make sure that we don't get any unwanted side effects in that case.

Corepex: ah nice i thought only for the first match - shitty CSS :P

Corepex: wrapped it in a div https://github.com/pimcore/studio-ui-bundle/pull/325/commits/819b265cf13197e379d4c84aa63fb09d73e993be#diff-d18079ee22fd1e056a3e26d670a4d749686da0dc4b25b2a52b29cb2d148a18cfR64

markus-moser:
- Can we please make the left margin of the table more like in the design so that it is in one line with the headline? 

image

vs.

image

- Can we please make the name and especially value column a bit wider so that longer values have enough space?

image

- Additonally I created a issue for implementing a client side sort approach: #341

@Corepex From my point of view we can merge this PR as soon as we solved the first two points 👍

PR #373, opened 2024-06-19
fixed: workflow place label styling logic

### LINKED ISSUE #345: [Workflows] Workflow place label styling logic
The workflow place configuration tree offers 2 styling/color related configuration option regarding workflow place labels:

- color
- colorInverted
https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/src/DependencyInjection/Configuration.php#L1457

**@studio-backend-team**: Please adjust the workflow detail API that it directly outputs those 2 options instead of the current options `backgroundColor`, `fontColor` and `borderColor`. Those values are calculated out of the original 2 configuration options by the workflow engine for the classic bundle UI. As we have a little bit different styling approach in the studio UI those values do not help anymore and we need the original config values to build or new login based on them in the UI.

@Corepex As soon as the backend team updated the API we should implement the logic for the place labels like discribed by @idaiv in this comment https://github.com/pimcore/studio-ui-bundle/pull/338#issuecomment-2165173791 based on the `color` and `colorInverted` options.  

image

Corepex:
Resolves #345 

## Additional info
I used the following table to ensure the opacity of the hex color: https://davidwalsh.name/hex-opacity


markus-moser (approved):
Works like a charm, let's merge it 👍

PR #375, opened 2024-06-19
Improved `schedule` column styling

### LINKED ISSUE #367: [Schedule] Display translation key instead of techical value in action column
* The dropdown shows nicely translated values whereas the column content is displayed without translation keys:

image

* The column value should not be centered but left aligned like in the design:

image

* Also the select styling (with purple color and arrow for the dropdown) for the action picker is not like in the design.

Related to #348

Corepex:
Resolves #367 
Depends on: https://github.com/pimcore/studio-backend-bundle/pull/156


markus-moser (approved):
Looks nice now, thanks!

PR #378, opened 2024-06-24
Redesigned schedule version picker

### LINKED ISSUE #368: [Schedule] Version picker
Version picker with just IDs will not work (it's too hard to find out the right version).

image

We also need to show the version date/time (with seconds) and version author like in the classic bundle:

image

Would it be possible to show all 3?
- version ID
- date + time (with seconds)
- author

Related to #348

Corepex:
Resolves #368


Corepex:
image

@markus-moser I redesigned the select accoding to the current design but i noticed that the column could get pretty wide based on the name of the user. I manually set the width to `220px` (see: https://github.com/pimcore/studio-ui-bundle/pull/378/files#diff-d4695a4234ba7692bfa7b7102fc9033f136787e14a8a2c891ea79fbfcecedffeR71). The problem is that the dropdown of the select is always as long as the cell ... so if the username would be more than lets say 10 chars it gets cut off ...

WDYT?

Corepex:
Finalized design from @idaiv https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=3021-140611&t=RjgKb95I2NTFB3uq-4

Corepex:
image

@markus-moser we have to check which options we have on paging the result ...

idaiv:
> image
> 
> @markus-moser we have to check which options we have on paging the result ...

Hi Chris & Markus! I would suggest to use the same pagination as for the trees: https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=1940-96199&t=RjgKb95I2NTFB3uq-4
This is the most narrowed and compact version, which makes it suitable for a dropdown. 
image
WDYT?

markus-moser:
@Corepex Isn't it be possible to just add a scrollbar in that case? Pagination for me sounds like quite a bit of effort?

Corepex:
that was a hard one :P i just switched back from `dropdown` to `select` and with some CSS magic everything works now as expected :D

markus-moser (changes requested):
@Corepex Looks pretty cool now, just a few more details:

- [ ] Please show the version count instead of the version ID
- [ ] Is it possible to make the version count bold like in the design?
image
- [ ] When a version is picked we should show the version count in the column, currently it looks like this:
image

markus-moser (approved):

PR #384, opened 2024-06-27
Added `breadcrumbs` and moved `tabbar` items

### LINKED ISSUE #346: Asset view breadcrumb and toolbar
Implement like in design in figma (next to each other):

image

Corepex:
Resolves #346


markus-moser (changes requested):
@Corepex This works already really good! This is what I found that we could improve:

- [x] In the design instead of showing the first folder/parent it shows the last one:
image
Can we switch this? The last folder is more important then the first one. I rellay like the approach with the `...` and the dropdown. 👍 
- [x] Clicking on the folder which is shown directly in the breadcrumb does not open the element. So in this case e.g. the "Car Images" folder:
image

markus-moser (approved):
nice, LGTM

PR #399, opened 2024-07-04
Asset preview

### LINKED ISSUE #75: Asset Preview
For different asset types
- image
- video
- documents ( = pdf preview)
- text files → see also Asset Text Editor

image

image

image

robertSt7:
partially resolves #75


markus-moser (changes requested):
The docs preview for PDF files still has a problem. In the classic bundle there is a mechanism to generate the preview PDF via the messenger queue. While the queue is running you will see a message like this then:

image

The same PDF was displayed like this in the studio UI:

image

robertSt7:
> The docs preview for PDF files still has a problem. In the classic bundle there is a mechanism to generate the preview PDF via the messenger queue. While the queue is running you will see a message like this then:
> 
> image
> 
> The same PDF was displayed like this in the studio UI:
> 
> image

@markus-moser This is not reproducible for me. Could you please describe some steps? Thank you

markus-moser:
@robertSt7 I'm not able to reproduce the wrong JSON view anymore too. But it seems that we still have the problem that this PDF preview queue logic is not respected. It's all about this `getResponseByScanStatus` in the classic bundle:

https://github.com/pimcore/admin-ui-classic-bundle/blob/7999ef435d84f133d469539b5664c6b1d65c176e/src/Controller/Admin/Asset/AssetController.php#L1436

There is a logic in place where the PDF preview generation is added to the messenger queue:

https://github.com/pimcore/admin-ui-classic-bundle/blob/7999ef435d84f133d469539b5664c6b1d65c176e/src/Controller/Admin/Asset/AssetController.php#L1468

And in the message handler it then checks if the PDF contains JS:

https://github.com/pimcore/pimcore/blob/a96a60244d7d9bb26a0d69500a7075efc6fac686/lib/Messenger/Handler/AssetUpdateTasksHandler.php#L67

We need to do somehow the same in the studio UI

assets/js/src/core/modules/asset/editor/types/document/tab-manager/tabs/preview/preview-container.tsx
```diff
@@ -11,20 +11,39 @@
...
   const assetContext = useContext(AssetContext)
-  const { data } = useGetAssetByIdQuery({ id: assetContext.id! })
+  const [docURL, setDocURL] = useState('')
+
+  useEffect(() => {
+    if (docURL !== '') {
+      return
+    }
+
+    fetch(`http://localhost/studio/api/assets/${assetContext.id!}/document/stream/pdf-preview`)
```
vin0401, line 28: Can we just use the relative path here? Without the domain?
Otherwise I would suggest that we create a new issue, to find out how we can make the domain configurable.

vin0401 (changes requested):
Just one small topic, we should take a more detailed look at. :)

assets/js/src/core/modules/asset/editor/types/document/tab-manager/tabs/preview/preview-container.tsx
robertSt7: see #453

PR #404, opened 2024-07-09
Added functionality to `cleanup all` button in `schedule` tab

### LINKED ISSUE #370: [Schedule] Cleanup all button
#369 needs to be done first

image

The "Clean up all" button for archived schedule entries is missing.

Related to #348

Corepex:
Resolves #370


markus-moser (changes requested):
Currently the clean up all button does not actually delete the archived scheduler entries but calls the asset versions delete call. Do we have an API for this? Are we able to use the single delete call for each entry and run multiple ajax requests for each single entry?

markus-moser (approved):
looks good now 👍

PR #406, opened 2024-07-11
removed `useTranslation` hook from helpers

Corepex:
## Additional info

... makes the hooks usable outside of components


robertSt7 (approved):

Corepex:
@markus-moser, @vin0401 

We have to find a solution for displaying the date independently from the local preferences fetched from `Intl`.

Example:

```ts
//Result: "20/12/2020, 04:23"
new Intl.DateTimeFormat('en-GB', {dateStyle: 'short', timeStyle: 'short'}).format(date)

//Result: "12/20/20, 4:23 AM"
new Intl.DateTimeFormat('en', {dateStyle: 'short', timeStyle: 'short'}).format(date)
```

We should somehow ensure that at least the `year` is always in a 4-digit format ...

A possible solution would be:
```ts
new Intl.DateTimeFormat('en', {
    month: 'numeric',
    year: 'numeric',
    day: 'numeric',
    hour: 'numeric',
    minute: 'numeric',
    second: 'numeric',
    hourCycle: 'h24'
  }).format(date)
  ```
  
  ... BUT that takes away the benefit to "quickly" style the dateTime with `dateStyle` and `timeStyle`.

> Hint: `dateStyle` and `timeStyle` cannot be combined with `year`, `month`, `hout` ...

PR #410, opened 2024-07-16
Add possibility to add properties in property tab

### LINKED ISSUE #329: [Properties] Implement add property functionality
- [ ]  Adding custom properties is not yet implemented. **The properties can be saved later together with the asset, so the add button should just add the new row to the table/form**:

image

- [ ]  Adding predefined properties not implemented yet (same like custom properties):
image

- [ ] When I edit the value of for example a text field the value get's not updated in the table (not sure if related to this PR or a "problem" of the inline edit functionality itself)

(follow-up to #324)

vin0401:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/329


markus-moser (approved):
I created some follow up issues but let's merge this one and tackle the other topics in separate PRs:

#412 
#413
#414 
#415

PR #423, opened 2024-07-22
Asset image sidebar

### LINKED ISSUE #214: Asset Image Sidebar
https://www.figma.com/file/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?type=design&node-id=272-70190&mode=design&t=3CiLs3iiBsnirWBR-4

image

- [ ] Details Info
- [ ] Thumbnail Download

robertSt7:
resolves #214


markus-moser:
@robertSt7 
It works already pretty good! I noticed a few problems:

- [x] Download of the "Original File" results in a broken file. Is this a UI or API topic?
- [x] Mode and format need to be selected otherwise the download results in a broken file. In the classic bundle those 2 values are prefilled with the default option. In my opinion we should do the same in the new UI. But when we do this we have the problem that we do not have labels for the fields and also the width of the values is a bit problematic:
image
@idaiv Could you please check what we can do here?

vin0401 (commented):
Codewise it looks good to me. :)

idaiv:
Hi @markus-moser!
I see your concerns, makes totally sense! My proposal would be to make simple select dropdowns with the label on top, the default selection in the dropdown input line and arrange these two one under each other. Here you can see it: https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=318-52086&t=IgcwTapm2nkQ8QQG-4 
WDYT?

markus-moser:
@idaiv Yes, thought about something similar - LGTM 👍

markus-moser:
@robertSt7 Please check the 2 open topics - let's solve it like in the design proposal from @idaiv. Afterwards we can merge it 👍

markus-moser:
@robertSt7 Thanks for solving it - checked your fixes and improvements and now all works well 👍

PR #407, opened 2024-07-11
Added `tags` tab

Corepex:
## Changes in this pull request
Resolves #160

PR #424, opened 2024-07-22
Improve tags tab

Corepex:
Followup: https://github.com/pimcore/studio-ui-bundle/pull/407
Related to: https://github.com/pimcore/studio-backend-bundle/pull/268


Corepex:
- [x] "Apply folder Tags" -> Assigns the tags of the direct parent to the current element (add - not replace!)
 - [x] "Remove current element Tags" -> Assigns the tags of the parent to the current element (replaces the element assign tags)
 - [x] add optimistic update (tag to table)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/tags/components/assigned-tags/assigned-tags.tsx
```diff
@@ -11,20 +11,41 @@
...
+import { useGlobalAssetContext } from '@Pimcore/modules/asset/hooks/use-global-asset-context'
 
 type TagWithActions = Tag & {
   actions: React.ReactNode
 }
 
 export const AssignedTagsTable = ({ tags, isLoading }: { tags: Tag[], isLoading: boolean }): React.JSX.Element => {
   const { t } = useTranslation()
+  const [loadingRows, setLoadingRows] = useState({})
+  const { context } = useGlobalAssetContext()
```
vin0401: Can we use the normal Asset Context in that case? 

The global context is set a bit later in the render cycle, than the normal asset context. 
That would save us some extra renders. :)

vin0401 (changes requested):
Just a small performance recommendation. :)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/tags/components/assigned-tags/assigned-tags.tsx
Corepex: I used the `useGlobalAssetContext` to also get the type since we could reuse it for other element types ...

vin0401: But the useGlobal**Asset**Context is also only used for assets. 

For sure we will have to add an additional context for other element types.
But that's something that I would see under a separate/new context.

vin0401: And for me it's something we should face in the moment when we start with the object area.

markus-moser (approved):
LGTM

PR #425, opened 2024-07-22
[Notes&Events] User is not displayed

### LINKED ISSUE #393: [Notes&Events] User is not displayed
In the classic UI bundle the user is visible in the `user` column.

image

In the new UI it's not visible. As far as I can see it's also not part of the design. 

@idaiv Could you please check the design and then hand over the issue to @robertSt7, thanks!

Related to #389

robertSt7:
resolves https://github.com/pimcore/studio-ui-bundle/issues/393


markus-moser:
Alternative approach included in #498

PR #430, opened 2024-07-23
[Versions] Add scroll bars & metadata

### LINKED ISSUE #352: [Versions] Asset metadata missing
The version preview should show asset metadata attributes:

image

Currently they are missing:
image

Related to #349

### LINKED ISSUE #419: [Versions] Scrolling
When the content is too long for the window size we need a scrollbar:

image

image

robertSt7:
resolves #419
resolves #352


markus-moser (changes requested):
Scrollbars work fine!

For the asset metadata I found some topics:

Compare the view from the classic bundle:

image

With the same version in studio UI:

image

For some data types there are some problems:
1. Date fields need to be formatted with the centralized date formatter instead of printing timestamps
2. Localized fields seem to not work correctly
3. Many to many relation fields seem to not work correctly. 

Not sure if 2. and 3. happens because we did not implement the logic for the asset metadata class definitions bundle yet? And also not sure if those kind of topics should be done in the UI world or on backend side? 

Another problem currently is that the version comparison does not work anymore in some cases. E.g. for asset ID 67 in the enterprise demo.

When I select one verison for the version preview javascript crashes and the tab looks like this:

image

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/versions/details-versions/details-versions-container.tsx
```diff
@@ -58,25 +58,46 @@ export const DetailsVersionsContainer = ({
...
         for (const key in data) {
           tempVersionData.push({
             [i18n.t('field')]: i18n.t(`version.${key}`)
           })
         }
 
+        for (const meta of metadata) {
+          data[`${meta.name} (${meta.type})`] = meta.data
+          tempVersionData.push({
+            [i18n.t('field')]: `${meta.name} (${meta.type})`
```
vin0401: We should use the `t`-function from `useTranslation` instead.

The normal i18n module has no reactivity inbuilt, this could cause problems in the future when the module is loaded before the translations are present.

vin0401 (changes requested):

PR #446, opened 2024-07-29
Added `custom-metadata` tab

### LINKED ISSUE #152: Custom Metadata (Community)
see https://github.com/pimcore/product-management/issues/377

https://www.figma.com/design/M5yXJeMQXI3HTdJldJWqCb/-Light-Theme--Complete-Templates?node-id=91-134928&m=dev

image

Allow detaching

Corepex:
Resolves #152


assets/js/src/core/components/grid/columns/types/language-select/language-cell.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { type DefaultCellProps } from '@Pimcore/components/grid/columns/default-cell'
+import React from 'react'
+import { SelectCell } from '@Pimcore/components/grid/columns/types/select/select-cell'
+import {
+  useGetSystemSettingsQuery
+} from '@Pimcore/modules/asset/editor/shared-tab-manager/tabs/custom-metadata/settings-slice.gen'
+
+export const LanguageCell = (props: DefaultCellProps): React.JSX.Element => {
+  const { data, isLoading } = useGetSystemSettingsQuery()
```
vin0401, line 22: hmm, I'm just thinking if it would be more efficient to centralize the settings, somehow. For example in the app loader. And we just then use a simple use hook to access all the data. I think this would give us a benefit compared to individual integrations in the different components. 

What do you think? :)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/custom-metadata/components/table/table.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  // const updateMetadata = (metadata: CustomMetadata): void => {
+  //  console.log('updateMetadata', metadata)
+  // }
+
+  function onUpdateCellData ({ rowIndex, columnId, value, rowData }): void {
+    console.log({ rowIndex, columnId, value, rowData })
+    // updateMetadata(value)
+  }
```
vin0401, line 45: I guess here is still the save missing?
Should we clean up a bit and mark it as todo?

vin0401 (changes requested):
Looks already pretty good code wise. :)

assets/js/src/core/components/grid/columns/types/language-select/language-cell.tsx
Corepex: i guess that makes sense - i could create a store and then we fetch it from there ...
i would do that in the followup :)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/custom-metadata/components/table/table.tsx
Corepex: this is just the implementation to already have the UI - no functions at all so yeah there is a save missing ;) - will be something for the followup if that's OK :100:

markus-moser (approved):

PR #456, opened 2024-08-07
add tooltips to editor tabs 

### LINKED ISSUE #435: [Asset Tabs] Add tooltips
Add tooltips to all tabs of the asset detail view:

image

The tooltip text (translation key) should be the same like for tab title.

xIrusux:
Add tooltips to editor tabs

Resolves #435 

## Additional info
Resorted to writing a custom icon wrapper to add and remove the "open" property on hover. 
Unfortunately Antd does not have an out of the box tooltip solution for the way we use their Tabs component (i.e. we remove the label text on click). 
I explored the option to add the tooltip on Tabs.TabPane but later [read that](https://github.com/ant-design/ant-design/issues/42021) it's (soon to be?) deprecated. 

Screenshot 2024-08-07 at 12 30 37


xIrusux:
https://github.com/user-attachments/assets/0e062069-bbf7-40ca-95db-d84d58037d34

small change to close tooltip on tab click incoming shortly

xIrusux:

xIrusux:
I don't love that I am manually setting the "openTabKey" to the first item in the array on load but have not found a smarter way to determine whats currently open. Open to ideas!

assets/js/src/core/components/editor-tabs/editor-tabs.icon-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+import React, { useState } from 'react'
+import { Tooltip } from 'antd'
+
+export interface IconWrapperProps {
+  tabKey: string
+  openTabKey: string | null
+  title: string
+  children: React.ReactNode
+}
+export const IconWrapper = ({ tabKey, openTabKey, title, children }: IconWrapperProps): React.JSX.Element => {
```
vin0401: Can we also add the tooltip, when the Tab has focus via keyboard? :)
Just want make sure that we meet the a11y requirements, if possible.

assets/js/src/core/components/editor-tabs/editor-tabs.tsx
```diff
@@ -30,11 +31,27 @@ export interface IEditorTabsProps {
...
   const { id } = useContext(AssetContext)
+  const [openTabKey, setOpenTabKey] = useState<string | null>(null)
+
+  useEffect(() => {
+    if (openTabKey === null && items.length > 0) {
+      setOpenTabKey(items[0].key)
+    }
+  }, [items])
 
+  const onTabClick = (key: string): void => {
```
vin0401: Does this also work with the keyboard? Otherwise we maybe could use the onChange listener instead?

assets/js/src/core/components/editor-tabs/editor-tabs.tsx
```diff
@@ -30,11 +31,27 @@ export interface IEditorTabsProps {
   showLabelIfActive?: boolean
 }
 
+export interface IconWrapperProps {
```
vin0401: I think this interface can be removed? Seems like it was moved to the tooltip component already? :)

vin0401 (changes requested):
Looks pretty good, already.
Just a few small adjustments. :)

assets/js/src/core/components/editor-tabs/editor-tabs.tsx
xIrusux: I had tested it and it also work for keyboard navigation. I do think the onChange handler seems more fitting though. Changed, Thanks!

assets/js/src/core/components/editor-tabs/editor-tabs.tsx
xIrusux: Absolutely!

assets/js/src/core/components/editor-tabs/editor-tabs.icon-wrapper.tsx
xIrusux: https://github.com/user-attachments/assets/ad318a2e-07aa-45ff-a823-4ab9649b2036

xIrusux: As I am not able to access each tab element separately (because of Tabs.TabPane soon being deprecated) 
I have made do with the <Tabs> onFocus/ onBlur combined with some event.target.id text matching. 

When someone is using the Keyboard to tab through the tabs and then switches to navigate via mouse, two tooltips can show at once which I think makes sense (see vid).

markus-moser: Yes, I'd agree that it makes sense to show 2 tooltips in that case...

vin0401 (approved):
Looks good! :)

Just one more organizational topic for the PRs. 
The PR title should always be written like a commit message, since it will be one after we squash it. ;D

Corepex:
@xIrusux, very nice first PR :1st_place_medal: nice to have you in the team :partying_face: :tada:

PR #457, opened 2024-08-07
Video sidebar

### LINKED ISSUE #80: Asset Video Sidebar
- Show video dimensions
- change video thumbnail for preview
- download video
- image preview → see Aset Video Image Preview

image

### LINKED ISSUE #81: Asset Video Image Preview
- Define image preview for video
- either via assigning an asset image
- or via selecting current player position

image

image

robertSt7:
resolves #80
resolves #81


assets/js/src/core/utils/route.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
vin0401: hmm, I think this file would be better located in the general app folder /app/api/pimcore.
For me it's something more essential for the app, rather than a utility to use when needed.

What do you think?

assets/js/src/core/modules/asset/editor/types/video/tab-manager/tabs/preview/preview-container.tsx
```diff
@@ -11,22 +11,57 @@
...
+    setPlayerPosition
+  }), [thumbnail, playerPosition])
 
   return (
-    <ContentToolbarSidebarView>
-      <PreviewView
-        src={ data!.fullPath! }
-      />
-    </ContentToolbarSidebarView>
+    <VideoContext.Provider value={ contextValue }>
```
vin0401, line 58: We could consider to move all context related code to a video provider component. That makes it a bit easier to reuse. :)

So far, when I interduce a new Context. I normally create a new Provider and a custom hook e.g. `useVideo`.
And it seems quite useful. In the hook you can easily switch to a new context in case we switch to another provider. And having the Provider abstracted makes it of course super reusable. :) 

What do you think?

vin0401 (changes requested):
Looks good already, I just have two small suggestions. :P

PR #459, opened 2024-08-08
426 emtpy screen notes events

### LINKED ISSUE #426: [Notes&Events] Implement empty screen
Please add a empty note the same way like already implemented for versions when the list of notes&events is empty:

image

xIrusux:

empty screen for notes and events

image

Resolves #426


markus-moser:
@xIrusux Please always add a link to the resolved issue to your PR like this, thanks:

Resolves #426

vin0401 (changes requested):
Code wise it looks good to me, but aren't we missing the button to add items to the list? :P

xIrusux:
:) 

https://github.com/user-attachments/assets/28f83b60-24f0-4591-9aaa-1bb9897acfe7

also tiny clean up of some bits as few styles are out of the box with flex anyway. Let me know if we rather want to be explicit. 

EDIT: made header bold so to match with most other headers we are not mega consistent here

markus-moser (approved):
Everything works like it should. The bold headline is correct. At the end the headline should be bold in all tabs. We should create a shared component for this, but let's create a separate issue for this.

markus-moser:
Separate issue for the headline topic #463

PR #473, opened 2024-08-19
Improve add custom property handling

markus-moser:
Fixed several issues regarding the add custom property handling:

* Focus key input field after clicking "add a custom property"
* Validate key and type as mandatory fields with modal error message
  * Reset values after clicking "cancel" to avoid validating outdated data
* Translate type select options


vin0401 (approved):
LGTM. :)

PR #474, opened 2024-08-20
Properties table improvements

markus-moser:
This PR improves various aspects of the properties table:

- The table is now 100% width by default and all columns are resizeable. 
     - All columns except the data column have a fixed with. 
     - The rest of the space is used by the data column
     - The data column has a min width of 300px
     - When resizing the data column the 100% width mode is disabled and all columns have there defined fixed width
     - The table autoWidth logic is now part of the grid component and can be re-used. It will work in combination with one single autoWidth column.
- Asset relations are displayed with filenames instead of just paths
- Drop target icon is now always right alligned
- Some margins/paddings/vertical alignments are fixed


assets/js/src/core/components/grid/grid.tsx
```diff
@@ -153,6 +157,48 @@ export const Grid = ({ enableMultipleRowSelection = false, modifiedCells = [], s
...
+    let autoWidthColumnFound: boolean = false
+    for (const column of columns) {
+      if (column.meta?.autoWidth === true) {
+        if (autoWidthColumnFound) {
+          throw new Error('Only one column can have autoWidth set to true')
+        }
+        autoWidthColumnFound = true
+      }
+    }
+  }
```
vin0401: We should wrap this in useMemo to only recompute it, when the columns change.
At the moment we do it on every render. Should not have to much effect but for the grid it can not hurt to save some computing time. :)

vin0401 (changes requested):
Just a small performance optimization. :)

assets/js/src/core/components/grid/grid.tsx
markus-moser: Thanks, implemented this and it seems to work!

PR #475, opened 2024-08-20
Improve add/remove custom asset metadata entry handling

markus-moser:
Similar approach like #473 for asset metadata.

Additionally it fixes the following:
- a metadata key can now be added mutliple times with different languages
- when removing a entry it only deletes the entry for the language of the row instead of all entries with the same name
- try to re-use and share type select option translation keys with properties


vin0401 (approved):
LGTM. :)

PR #477, opened 2024-08-20
Dependencies table autoWidth and scrolling

markus-moser:
The dependencies tables are now resizeable and 100% width too. Additionally I improved the scrolling behaviour when there is not enough space for various cases.


vin0401 (approved):
Looks good. :)

PR #462, opened 2024-08-12
Optimized properties table when showing `all` properties

Corepex:
## Changes in this pull request
Resolves #331 
Blocked by: #461

PR #478, opened 2024-08-21
Improved readonly view of `properties`

### LINKED ISSUE #331: [Properties] "All properties" needs to be a read only view
The "All Properties" tab is a readonly view. Therefore the inline edit functionallity needs to be disabled and the "Delete" icon should be hidden.

(follow-up to #324)

Corepex:
Resolves #331 
Followup to: #462


markus-moser (approved):
cool, it works 👍

markus-moser:
Went for an alternative approach in #507 -> closed

PR #479, opened 2024-08-21
463 headline shared component

### LINKED ISSUE #463: Unify tabs headline styling
* In most of the tabs the headline is bold but e.g. in the dependencies tab it's not (Requires + required by)
* The space above the headline is not the same for all tabs - this should be unified, so that when tabs are switched the headline stays at the same position.
* The headline should be a shared component so that it's reusable accross the tabs.

```[tasklist]
### Tasks
- [ ] Add another shared component to unify outer most container inside the tabs (i.e. padding)
```

xIrusux:
Resolves #463 

- added shared headline component 
- added and used shared card component 
- only tab that does not use the above is tags tab as it has a different layout. (Which I can also adjust to - just wanted to get an opinion first)

## Additional info

Before:

https://github.com/user-attachments/assets/1a7df97b-b794-45f5-9d86-2a67b4533fa0

After:

https://github.com/user-attachments/assets/08c17a53-9e15-40e5-be39-6d68cdd0c8df

Everything jumps less but I really hope I did not interfere with any important styling. 
As I have shifted quite a few things in the process / removed obsolete styling


markus-moser:
Beautiful!

xIrusux:
Thanks @markus-moser!
Lots of interfering styles and fiddling :D but worth it in the end.

xIrusux:
from conversation with @vin0401 

- [x]  rename to content-padding-container xxx  / content header 
- [x] - 'tab content containers' folder for both 
- [x] - use flex ant d in internal containers 
- [x] - use title component 
- [x] - write story book

xIrusux:
removed pagination from notes and events tab when there are no notes and events to remove some more jumpiness around the "no content" view.

I will implement the new button components in the different places the tab header uses them as a next step as part of issue: #500

How we differentiate between the content container you created and the content-padding-container and what belongs around that will also need some more thought but I'd also treat that as a separate concern so we can close this one.

vin0401 (approved):

PR #480, opened 2024-08-21
Asset metadata table improvements

markus-moser:
Various fixes for the asset metadata table:
- Fix language select cell
- Implement textarea type
- 100% width + resizeable columns


assets/js/src/core/components/grid/columns/types/textarea/textarea-cell.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { respectLineBreak } from '@Pimcore/utils/helpers'
+
+export type DefaultCellProps = CellContext<any, any>
+
+export const TextareaCell = (props: DefaultCellProps): React.JSX.Element => {
+  const { isInEditMode, disableEditMode, fireOnUpdateCellDataEvent } = useEditMode(props)
+  const { styles } = useStyle()
+  const [textAreaValue, setTextAreaValue] = useState(String(props.getValue() ?? ''))
+
+  // useEffect(() => {
```
vin0401: I think we can remove that comment? :) 

But now, where I see the useEffect. Do we focus the textarea, the moment we enter the editmode?
Otherwise I would prefer that we still add this. It saves a click. :D

assets/js/src/core/components/grid/columns/types/textarea/textarea-cell.tsx
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+import { type CellContext } from '@tanstack/react-table'
+import React, { useState } from 'react'
+import { useEditMode } from '../../../edit-mode/use-edit-mode'
+import { useStyle } from './textarea-cell.styles'
+import TextArea from 'antd/es/input/TextArea'
+import { respectLineBreak } from '@Pimcore/utils/helpers'
+
+export type DefaultCellProps = CellContext<any, any>
```
vin0401: I would prefer to use the type from the DefaultCell: https://github.com/pimcore/studio-ui-bundle/blob/1.x/assets/js/src/core/components/grid/columns/default-cell.tsx#L27

Otherwise we miss some type completion. :)

vin0401 (changes requested):
Just two small improvements.

assets/js/src/core/components/grid/columns/types/textarea/textarea-cell.tsx
markus-moser: Thanks, the focus is a good point - I fixed this!

assets/js/src/core/components/grid/columns/types/textarea/textarea-cell.tsx
markus-moser: Copied this from the text-cell -> I changed it now in both components

PR #481, opened 2024-08-21
Sort embedded metadata by name

markus-moser:
It's now possible to define a default sort order for the client side grid sorting. If the default sort order is defined it's not possible to disable sorting anymore (unsorted state does not make sense e.g. in the asset metadata table)


vin0401 (approved):
Nice! :)

PR #489, opened 2024-08-26
allow edit key in properties

### LINKED ISSUE #421: [Properties] Make key editable
Once a property is added the key is currently not editable:

image

xIrusux:
Resolves #421

## Additional info

- There is no primary key being used on the properties or custom meta date table. Hence changing the key requires a update of all properties. Let's discuss background 

These "full updates" have several effects: 
- one can not put a little tag on the changed cells like in the old ui which seemed useful
- when there are a lot of properties, redux will get very busy unnecessarily 
- one can edit a second property to have the same key as a first property which leads to all kind of update problems (I will ensure this does not happen via a modal in the ui though in a follow up commit)


xIrusux:
I went with using the old key passed in to only update the specific row

Am now also ensuring the data cant be published when key (or name for metaData) are empty or the same as another key/ name. 

Have done a quickfix of the modified data flag but it will need a bit more looking into to work fully with the new editable field. 

Not yet ready to review

xIrusux:
https://github.com/user-attachments/assets/4ba3d224-7442-4ab0-9ba4-3f244ed4aaf8

input controls: 
- mandatory field modal 
- duplication not allowed modal

xIrusux:
modified flag now showing when changes were made

https://github.com/user-attachments/assets/da762bdf-2ca8-459b-968d-e968916a59e8

I have noticed that the check box does not function. i.e. the saving functionality does not work. 

'checkbox' type cells 
checkbox-cell.tsx / boolean-cell.tsx / the way the checkbox is ingested in properties/components/table.table.tsx needs to be reduced down to one option and then have the editable functionality added + ensure the "modified" flag shows when a change happens. -> started to spike but now created Issue: #494 to also wait for markus's changes

assets/js/src/core/components/grid/grid.tsx
```diff
@@ -293,7 +293,7 @@ export const Grid = ({ enableMultipleRowSelection = false, modifiedCells = [], s
   ), [table, modifiedCells, data, columns, rowSelection, internalSorting])
 
   function getModifiedRow (rowIndex: string | number): GridProps['modifiedCells'] {
-    return memoModifiedCells.filter(({ rowIndex: rIndex }) => rIndex === rowIndex) ?? []
+    return memoModifiedCells.filter(({ rowIndex: rIndex }) => Number(rIndex) === Number(rowIndex)) ?? []
```
vin0401: the rowIndex can be a real text index or a number. 
I think a hard cast to Number does not seem to be an option here.

vin0401 (approved):
Looks good already, but still there is one change we have to face here.

assets/js/src/core/components/grid/grid.tsx
xIrusux: cool thanks. 
I will take another look how I can solve it :)

assets/js/src/core/components/grid/grid.tsx
```diff
@@ -292,7 +292,7 @@ export const Grid = ({ enableMultipleRowSelection = false, modifiedCells = [], s
     </div>
   ), [table, modifiedCells, data, columns, rowSelection, internalSorting])
 
-  function getModifiedRow (rowIndex: string | number): GridProps['modifiedCells'] {
+  function getModifiedRow (rowIndex: string): GridProps['modifiedCells'] {
     return memoModifiedCells.filter(({ rowIndex: rIndex }) => String(rIndex) === rowIndex) ?? []
   }
```
xIrusux, line 297: the incoming rowIndex ist typed to always be a string, while the rIndex has been a number in the cases I have seen. 
String() will convert it so its types are comparable in any case

PR #490, opened 2024-08-26
Introduce asset metadata type provider and refactor tab specific cell types

### LINKED ISSUE #352: [Versions] Asset metadata missing

### LINKED ISSUE #476: Custom asset metadata types extendable?
In the old UI it is possible to register custom asset metadata types via ExtJS so that they appear in the select list:

image

In the new UI the type list is hardcoded. Do we need to provide the possibility to add custom types there (in addition to the asset metadata class definitions...)?

markus-moser:
* Resolves #476
* Resolves #352 
* Improves the version preview and comparsion
* Still needs improved UI components but works relatively good

Related to https://github.com/pimcore/studio-backend-bundle/issues/393


vin0401 (approved):
Really like the new approach. 👍

PR #519, opened 2024-09-04
Cleanup asset tabs

markus-moser:
- Check which tabs should be detachable
- Check which tabs should be displayed in the various asset types
- Centralized shared tab manager definitions
- Add archive asset type
- Move text editor tab to text type


vin0401 (approved):
Nice, Finally the tabs are completely centralized. :)

PR #523, opened 2024-09-05
refactor version to accordion shared component 

xIrusux:
Refactor to use shared component version


vin0401 (approved):
As discussed we should check if we can make it possible to use the accordion/collapse component or at least a variant of it. Also we should work on the naming to make it a bit more generic and therefor suitable for other cases/features apart from versions.

But you get the approval to not block your progress. :)

PR #527, opened 2024-09-09
Added Asset Upload via DnD

### LINKED ISSUE #257: Asset Upload
- Single Upload
- Multi Upload
- ZIP Upload
- Upload via d&d 
- use browser defaults as far as possible

Corepex:
Resolves #257


assets/js/src/core/components/tree/components/context-menu/context-menu.tsx
```diff
@@ -71,7 +70,7 @@ export const TreeContextMenu = ({ children, node }: TreeContextMenuProps): React
 
           if (formData.has('file')) {
             // @ts-expect-error - marvin will know why
-            await dispatch(assetUpload.initiate({ body: formData, parentId: parseInt(node.id) })).unwrap()
+            await dispatch(assetUpload.initiate({ body: formData, parentId: parseInt(node.id) }))
```
vin0401: We should move the asset related code out of the tree component, because it's not directly related to UI components.

In general, there should be no more business logic be added under the components folder,
instead we should locate this specific tree logic somewhere close to modules/asset/tree.

assets/js/src/core/components/tree/components/context-menu/context-menu.tsx
```diff
@@ -94,7 +93,7 @@ export const TreeContextMenu = ({ children, node }: TreeContextMenuProps): React
 
           if (formData.has('zipFile')) {
             // @ts-expect-error - marvin will know why
-            await dispatch(archiveUpload.initiate({ body: formData, parentId: parseInt(node.id) })).unwrap()
+            await dispatch(archiveUpload.initiate({ body: formData, parentId: parseInt(node.id) }))
```
vin0401: Same here, let's move it to modules, somehow. :)

assets/js/src/core/components/tree/node/tree-node.tsx
```diff
@@ -153,8 +155,30 @@ const TreeNode = ({
...
+  const uploadProps: UploadProps = {
+    action: `/studio/api/assets/add/${id}`,
+    name: 'file',
+    multiple: true,
+    openFileDialogOnClick: false,
+    showUploadList: false,
+    onChange: ({ fileList }) => {
+      setUploadFileList(fileList.filter((file) => file.status === 'uploading'))
+    }
+  }
```
vin0401: Also not UI relevant should be moved to the asset tree under modules.

assets/js/src/core/components/tree/node/tree-node.tsx
```diff
@@ -153,8 +155,30 @@ const TreeNode = ({
...
+  function onDragOver (event: MouseEvent): void {
+    const assetMetaData = metaData?.asset
+
+    if (assetMetaData !== undefined && assetMetaData.type === 'folder') {
+      setSelectedIds([id])
+    }
+  }
+
+  const uploadProps: UploadProps = {
+    action: `/studio/api/assets/add/${id}`,
```
vin0401, line 177: Do we have an alternative here? 

Would be nice if we could use the normal api endpoints provided by RTK Query.
Otherwise we have to find a way to invalidate the cache tag right after the successful upload.

assets/js/src/core/components/upload/upload-list/upload-list.tsx
```diff
@@ -0,0 +1,73 @@
+/**
```
vin0401, line 1: I think we can create a story for this? 
Or is the behavior of the upload still open for discussion?

vin0401 (changes requested):
Looks quite good already. But I would love to see a stronger separation here. Otherwise it could be quite hard to reuse the tree component for other features like objects or the bookmarklist, etc.

vin0401:
Ohh, there is one last thing I forgot to mention. When we drag a file somewhere else than the drop target it replaces the browser url and shows the image. I think we should prevent this behavior, because it can happen that you miss an area by dragging to fast etc.

assets/js/src/core/components/tree/components/context-menu/context-menu.tsx
Corepex: done with https://github.com/pimcore/studio-ui-bundle/pull/527/commits/7ac29efdc39882e0150aa4f8cc710aaa6b66fc4c

assets/js/src/core/components/tree/components/context-menu/context-menu.tsx
Corepex: done with https://github.com/pimcore/studio-ui-bundle/pull/527/commits/7ac29efdc39882e0150aa4f8cc710aaa6b66fc4c

assets/js/src/core/components/tree/node/tree-node.tsx
Corepex: done with https://github.com/pimcore/studio-ui-bundle/pull/527/commits/7ac29efdc39882e0150aa4f8cc710aaa6b66fc4c

assets/js/src/core/components/tree/node/tree-node.tsx
Corepex: no :D as we both tried it there is no alternative but we found a way :D

vin0401 (approved):
Works like a charm. :P

PR #528, opened 2024-09-09
Refactor to accordion shared component / notes & events

xIrusux:
Resolves #


xIrusux:
**Storybook Accordion Options:**

Default - The Chevron Button is custom to allow a longer subtitle than title, this was not possible in Ant Collapse

Chevron Left 

Collapse Disabled - here one has to add 'collapsible:'icon'' to the disabled item. This disables the Ant D functionality to expand on header click - and I have piggy backed on it to hide the custom Chevron. I think the wording 'icon' is misleading and could extend the prop and rename. This would however mean its harder to know whats under the hood and look this up. Please let me know opinions if you have any :)

Spaced - as used in notes and events. Custom styling with space between collapse panels. 

Active Key Controlled - as used in version. This allows controlling nested Accordions and allowing to only open one panel at a time

ExclusiveAccordion - Allows exclusively opening one of several Accordion Panels

Let me know if you have any better naming ides :)) 

https://github.com/user-attachments/assets/9a14bfa3-b6de-4570-b734-41cac0340209

Video of current refactor including some feature updates:

Versions Accordion is now an accordion in the sense that only one version collapse can be open at a time. 
Except for when we are comparing -> then there are 2 open version collapse possible at any one time. 
This did now work before hand.

assets/js/src/core/components/accordion/accordion.tsx
```diff
@@ -11,24 +11,117 @@
...
 }
 
-export const Accordion = ({ items, exclusive = true, ...props }: AccordionProps): React.JSX.Element => {
+export const Accordion = ({
+  items,
+  accordion = false,
+  spaced = false,
+  className,
+  activeKey,
+  alignChevronLeft = false,
```
vin0401: I would reuse or overwrite the api from ant here. 

https://ant.design/components/collapse#collapse
expandIconPosition => it can be start or end. 

Maybe we start with `start` and `after-title`  for now.
And when we have to add something new we can figure out how to do it.

The boolean value would limit us to only two positions.

assets/js/src/core/components/version-accordion/version-accordion.tsx
```diff
@@ -68,19 +68,20 @@ export const VersionAccordion = ({
```
vin0401: I'm still struggling with this component. How do we reuse it with versions in the name?
And props tailored so specific for versions?

I think we need a new component for the versions. Something like TimelineAccordions.
API vise we could add props for items (List combined out of our new accordion component and the necessary timeline props), selectedKeys, selectable, etc.

assets/js/src/core/components/accordion/accordion.tsx
```diff
@@ -11,24 +11,117 @@
...
-import React from 'react'
-import { Collapse, type CollapseProps } from 'antd'
+import React, { useEffect, useState } from 'react'
+import { Collapse, type CollapseProps, Flex } from 'antd'
 import { useStyles } from '@Pimcore/components/accordion/accordion.styles'
+import { type ItemType } from 'rc-collapse/es/interface'
+import i18n from 'i18next'
+import { IconButton } from '@Pimcore/components/icon-button/icon-button'
+
+export interface AccordionItemType extends ItemType {
```
vin0401, line 26: Maybe we should also add a theme property? :)

Like success (the greenish one we have for versions), primary (the purple one we have for versions) and for sure some other types. That would make it quite easy to reuse them through the whole app. :)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/notes-and-events/notes-and-events-view.style.tsx
```diff
@@ -22,12 +22,50 @@ export const useStyles = createStyles(({ token, css }) => {
...
+      }
+
+      .ant-table-wrapper p {
+        margin: 0;
+      }
+
+      .ant-table-wrapper .ant-table-cell {
+        white-space: unset;
+        overflow-wrap: anywhere;
+      }
```
vin0401, line 67: We have a lot of new styles here. Why do we need them?

vin0401 (commented):
We are heading in the right direction. :)

assets/js/src/core/components/accordion/accordion.tsx
xIrusux: Yes, I wasn't sure if that would be to specific but will add that in now too :)

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/notes-and-events/notes-and-events-view.style.tsx
xIrusux: @vin0401 They have been moved from the notes-and-events-card.style and refer to the part of the code where we are able to display a grid if there is note.data defined. 
I have not been sure how to create that scenario. Would you know how?

Screenshot 2024-09-11 at 12 28 38

assets/js/src/core/components/accordion/accordion.tsx
xIrusux: That makes sense. I will give that a go.

assets/js/src/core/components/version-accordion/version-accordion.tsx
xIrusux: @vin0401 ok - I saw this more as just a refactoring of what was there and not as a new shared component to use. 
I thought this is to big and specific of a component to be made shareable? Likely I am just missing the full picture here and dont understand why that would be useful.

assets/js/src/core/components/accordion/accordion.tsx
xIrusux: I will use the same overwrite logic for the "collapse="icon"" property that is used in items atm to disable collapsing. 

Edit: 
has been replaced by disabled prop

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/notes-and-events/notes-and-events-view.style.tsx
vin0401: okay, I think we can leave it like that for now. And hopefully we can remove it, the moment we have some of the spacing components.

assets/js/src/core/components/version-accordion/version-accordion.tsx
vin0401: In general we are not really limited in the size of the components. I would see that as a typical organism in regards to atomic design. 

Making something reusable in regards of UI is never a bad approach. Maybe we don't have a reusable case, yet. But maybe it will come up in one of our bundles or the community will find a practical use for it. 

And is it really that specific? Just imagine it without the content. It's just a timeline that takes accordions as a child.
How the content will look like in the end should be something that is up for the developer who uses the component. Content he could plug & play together out of other atoms/molecules/organism.  

Hope that helps? Otherwise you can call me anytime :P

assets/js/src/core/components/accordion/accordion.tsx
xIrusux: Done :)

xIrusux:
Screenshot 2024-09-11 at 17 20 19

Here we go :)

assets/js/src/core/components/accordion/accordion.tsx
xIrusux: Done :)

assets/js/src/core/components/version-accordion/version-accordion.tsx
xIrusux: can we merge this one and I open a new pull request for this ? :)

assets/js/src/core/components/accordion/accordion.styles.ts
```diff
@@ -14,25 +14,126 @@
...
+    highlightBackgroundColor: '#F6FFED',
+    highlightBorderColor: '#B7EB8F',
+    highlightColor: '#52C41A',
+    signalBackgroundColor: '#E6F4FF',
+    signalBorderColor: '#91CAFF',
+    signalColor: '#1677FF',
+    ...token
+  }
+
   return {
```
vin0401, line 27: We should follow the BEM naming here. 

For example is the chance of having a success class in the code quite high.
With following BEM we can make sure that we don't get any unwanted side effects.

vin0401 (changes requested):
One last thing

assets/js/src/core/components/version-accordion/version-accordion.tsx
vin0401: Yeah sure, just marked one last small thing. :) Can you create an issue for this topic? :)

assets/js/src/core/components/accordion/accordion.styles.ts
xIrusux: Good point - I gave this a go int he last commit :) @vin0401

vin0401 (approved):

xIrusux:
The PR now also includes a new accordion timeline shared component + an updated vertical timeline shared component
Screenshot 2024-09-19 at 11 56 52

Screenshot 2024-09-19 at 10 10 31

It has also been refactored to BEM

vin0401 (approved):
There are still some linting issues open, but apart from that it looks good. :)

PR #538, opened 2024-09-19
fix wrong classname

bramstroker:
Resolves #


markus-moser:
Thanks a lot for your PR.

It also helped us to identify that we have some problems with the Github pipelines when we receive PRs from forks. We where able to fix this and pipelines are working now (see https://github.com/pimcore/studio-ui-bundle/pull/566). For this PR we would need a upstream merge to fix them but that's not worth the effort. Therefore I merge it allthough pipelines failed.

PR #541, opened 2024-09-19
539 switch shared component

### LINKED ISSUE #539: Make Switch shared component

xIrusux:
Resolves #539 

## Additional info
Screenshot 2024-09-19 at 17 52 45
Screenshot 2024-09-19 at 17 52 45


assets/js/src/core/components/switch/switch.tsx
```diff
@@ -0,0 +1,43 @@
...
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import React from 'react'
+import type { SwitchProps as AntdSwitchProps } from 'antd/es/switch/index'
+import { Switch as AntdSwitch } from 'antd'
+import { Flex } from '@Pimcore/components/flex/flex'
+
+export interface SwitchProps extends AntdSwitchProps {
+  tag: string
```
vin0401: I would rather call it a label. When I read tag, I have to think of our tag component.

Have you checked if this is maybe also already possible with the Form/FormItem component from ant? 
It would be good to have a story example for that to. Especially with the placement of the label.

vin0401 (changes requested):

assets/js/src/core/components/switch/switch.tsx
xIrusux: Defining where the label sits via a form component is not possible but I have created a wrapper and added a story for it. 
Currently the wrapper does not do a whole lot though . Let me know what you think.

vin0401 (approved):

PR #542, opened 2024-09-19
536 shared component tags

### LINKED ISSUE #536: Refactor Tags in Version to no longer require styling and instead use AntD IconTags with the according themes
Themed Tags from Ant to Use:
Screenshot 2024-09-19 at 11 44 02

What to replace:
Screenshot 2024-09-19 at 11 46 19
Screenshot 2024-09-19 at 11 47 04

xIrusux:
Resolves #536


assets/js/src/core/components/tag/tag.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import { Tag as AntTag, type TagProps as AntTagPropsProps } from 'antd'
+import React from 'react'
+import { useStyles } from './tag.styles'
+import { Icon } from '@Pimcore/components/icon/icon'
+import { type LiteralUnion } from 'antd/es/_util/type'
+import type { PresetColorType, PresetStatusColorType } from 'antd/es/_util/colors'
+
+export interface TagProps extends AntTagPropsProps {
+  text: string
```
vin0401: Why not use the default behavior from ant and just use the children instead of a new text property?

assets/js/src/core/components/tag/tag.tsx
```diff
@@ -0,0 +1,97 @@
...
+  iconName?: IconNameType
```
vin0401: hmm, why do we have this limitation of icons?
I would rather accept any string. Otherwise the tag component is not really flexible.

assets/js/src/core/components/tag/tag.tsx
```diff
@@ -0,0 +1,97 @@
...
+  className?: string
+  theme?: TagTheme
+}
+
+export type IconNameType = 'world' | 'user'
+
+export type TagTheme = 'transparent' | 'user-role' | 'admin-role' | 'link-blue' | 'link-purple'
```
vin0401: hmm, for me user-role and admin-role do not really sound like themes. 
I would rather see them as own components, that will implement the tag.

Same for link-blue and link-purple. Mixing the names of the color in here, could create conflicts in the future, when we switch to other colors. And again we are not really adding new functionality, so maybe we can directly use the color and bordered property for it? But we also could think of adding a LinkTag-Component, when you think it would be helpful.

assets/js/src/core/components/tag/tag.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+        &.ant-tag.theme-transparent {
+            background-color: transparent;
+        }
+
+        &.ant-tag.theme-user-role,
+        &.ant-tag.theme-admin-role {
+            background-color: rgba(0, 0, 0, 0.04);
+            color: #565656;
+            border-color: #EBEBEB;
+        }
```
vin0401: Do they really differ from the default color variant?

vin0401 (changes requested):

assets/js/src/core/components/tag/tag.tsx
xIrusux: It felt more concise to me as I thought it would be unlikely that anything other than an actual string would be used here. Is this so the benefit is that someone can pass in a react element? 
I have changed it :)

assets/js/src/core/components/tag/tag.tsx
xIrusux: @vin0401 
My thought behind it was that it would be good to have some kind of feedback when an icon name does not yet exist. 
I played a bit with using the IconLibrary to determine which strings are allowed but Storybook does not like me using the injection in this way :P 

Screenshot 2024-09-23 at 13 33 49

Have made it a n open string in coming push. Maybe good enough for now.

assets/js/src/core/components/tag/tag.styles.ts
xIrusux: @vin0401 This depends if you want to take the story book as the be all and end all. 
The colors are definitely different. 
I spoke about this with ida last week and have just followed up to get an idea how accurate to be going forward.

xIrusux: there is a difference between the AntD default theme and our default style theme. 
To use the "pimcore custom" default one has to add the color "default" explicitly when using the Tag. 
i.e. <Tag color="default">. 
Let me know what you think. 

---

edit: replaced all of that by just using the AntD token magic

assets/js/src/core/components/tag/tag.tsx
xIrusux: Have improved this. 
Let me know what you think about the way I listed the different variations in Storybook.

vin0401 (approved):
Looks good to me. :)

PR #543, opened 2024-09-20
Update README.md

ValeriaMaltseva:


ValeriaMaltseva:

README.md
```diff
@@ -1,7 +1,13 @@
-# Studio Ui
+# Studio UI
 
 The Studio UI Bundle provides a Backend UI for Pimcore. It is based on the [React](https://react.dev/) framework.
 
+#### Backend Admin Panel: `{your-domain}/admin`
+
+#### Frontend UI: `{your-domain}/pimcore-studio`
```
vin0401: I think we don't need the current admin domain here anymore.

This should be enough. :)
Admin UI: {your-domain}/pimcore-studio 

Also this step is not needed anymore:
- Login via the classic backend UI

Could you remove it? :) Christian already implemented the authentication for studio. So there is no need to use the current admin for login anymore.

vin0401 (changes requested):

vin0401 (approved):
Looks good. :)

PR #545, opened 2024-09-20
Prepare data object context and tabs

markus-moser:
Implement general data object context and element context handling + the following data object tabs:
- Properties
- Schedule
- Dependencies
- Notes & Events
- Tags
- Workflow
- Versions Tab preparation - dummy components for single and comparison view are used for now


assets/js/src/core/app/config/services/index.ts
```diff
@@ -73,6 +79,12 @@ container.bind(serviceIds['Asset/Editor/ArchiveTabManager']).to(ArchiveTabManage
 container.bind(serviceIds['Asset/Editor/UnknownTabManager']).to(UnknownTabManager).inSingletonScope()
 container.bind(serviceIds['Asset/MetadataTypeProvider/MetadataTypeRegistry']).to(MetadataTypeRegistry).inSingletonScope()
 
+// Data Objects
+container.bind(serviceIds['DataObject/Editor/ComponentRegistry']).to(ComponentRegistryService).inSingletonScope()
```
vin0401, line 83: To be honest I'm not sure why this one was bound initially to the assets. 
It should be generic for all case.

vin0401 (changes requested):
Just one small thing with the component registry. 
Apart from that it's super awesome to see the first steps for the objects! :)

assets/js/src/core/app/config/services/index.ts
markus-moser: Should it be in the element context or totally global for all kind of components?

vin0401: I would make it totally global. So far I would not see the benefit to limit it to a specific area. 
The components are registered via a descriptive identifier, anyway.

markus-moser: Fixed in #563

Issue #544, opened 2024-09-20
Improve UX of split layout resizer


vin0401:
The current version of our split-layout needs a bit of an improvement in terms of UX.

The resizer is hard to use and provides to less user information in regards to the interactivity.
The improved behavior is noted down here:

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=6082-240242&node-type=canvas&t=3CCqdBd2pVVmoxO2-0

ValeriaMaltseva:
Agreed that the user should always see the resizer button, even if this functionality isn’t available (when the section width is less than the minimum width).

PR #546, opened 2024-09-23
[544] Improve UX of split layout resizer

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/544

Resolves #


assets/js/src/core/components/split-layout/components/divider/divider.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+      {isButtonVisible && (
+        <div className={ styles.iconContainer }>
+          <Icon
+            className={ styles.icon }
+            name={ 'chevron-selector-horizontal' }
+            options={ { height: 14, width: 14 } }
+          />
+        </div>
+      )}
```
vin0401: hmm, I'm wondering if we could use our IconButton Component in that case?
Because it already looks pretty similar. 

Maybe we just have to introduce a new type variant, in case we can't match it with the current options.

assets/js/src/core/components/split-layout/components/divider/divider.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  return (
+    <div
+      className={ cn(styles.dividerContainer, { [styles.resizable]: isResizable }) }
+      onMouseDown={ handleMouseDown }
+      onMouseEnter={ handleMouseEnter }
+      onMouseLeave={ handleMouseLeave }
+      ref={ dividerRef }
+      role={ 'button' }
+      tabIndex={ -1 }
```
vin0401: Do we need this tab index? 
There are no keyboard events bound to it. So for what is it used?

vin0401 (changes requested):
Looks pretty good already. :)

assets/js/src/core/components/split-layout/components/divider/divider.tsx
ValeriaMaltseva: @vin0401 Deleted this code and added ESLint disable rules instead.

assets/js/src/core/components/split-layout/components/divider/divider.tsx
ValeriaMaltseva: @vin0401
Firstly, I explored using the IconButton component. However, apart from the visual styling, I noticed some differences in behavior. In my case, I only need to change the cursor type, whereas the IconButton modifies both the background color and icon color. For this reason, it feels like an edge case to me.

Moreover, using the IconButton would require altering existing styles, which could lead to re-checking the component in areas where it’s already implemented.

What are your thoughts?

vin0401: Do we really need an altering on existing styles? 

I mean we also could introduce a new theme or type.
That should not have much effect on existing styles.

image

And maybe we also should add a new sizing in case we can't match that yet.

vin0401 (approved):

PR #555, opened 2024-09-26
Batch edit modal

xIrusux:
Resolves #


assets/js/src/core/components/modal/footer/modal-footer.tsx
```diff
@@ -13,22 +13,31 @@
 
 import React from 'react'
 import { useStyle } from '@Pimcore/components/modal/footer/modal-footer.styles'
+import { Flex } from '@Pimcore/components/flex/flex'
 
 interface IModalFooterProps {
   children?: React.ReactNode
   buttonAlignment?: 'space-between' | 'end'
```
vin0401: I think we can improve this a bit more. I think we used that before we had our flex component.
We could define the justify prop from flex as prop here.

vin0401 (changes requested):
We are on the right way. Just one small thing, I would like to improve. :)

assets/js/src/core/components/modal/footer/modal-footer.tsx
xIrusux: Ah! I was meant to be using the shared component flex container! 
Good spot :)

xIrusux: @vin0401 fixed :)

vin0401 (approved):

Issue #550, opened 2024-09-24
Wrapping Tag Shared Component


xIrusux:
https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=5594-229685&m=dev

![Image](https://github.com/user-attachments/assets/ca2fba1a-d0e4-4953-98cb-4d55f75a367f)

vin0401:
I think we could create an own component for the divider used under the tags. 
Maybe it also makes sense to handle that in a own issue.

idaiv:
Hi all @ValeriaMaltseva and @vin0401! As we discussed in our call, certain adjustments were needed and some more written information on the functionality. There is a dedicated Figma page now for this Tag Group Accordion here: https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=6380-233277

Please take a look at the notes and feel free to reach out!

PR #562, opened 2024-10-01
Wrapping Tag Shared Component

ValeriaMaltseva:
Task link: https://github.com/pimcore/studio-ui-bundle/issues/550

Resolves #

## Additional info

Example of using:
```
<TagList
  itemGap={ { x: 'small', y: 'small' } }
  list={ [
    [
      { children: 'Marvin Bertram', iconName: 'user-01', color: 'default' },
      {
        children: (
          <Text
            ellipsis={ { tooltip: { title: 'Martin Alexander Feldkirchner', placement: 'top' } } }
            style={ { width: '100px' } }
          >
            Martin Alexander Feldkirchner
          </Text>
        ),
        iconName: 'user-01',
        color: 'default'
      },
      { children: 'Ida Ivanova', iconName: 'user-01', color: 'default' },
      { children: 'Niklas Purnhagen', iconName: 'user-01', color: 'default' },
      { children: 'Christian Fasching', iconName: 'user-01', color: 'default' },
      { children: 'Christine Altmann', iconName: 'user-01', color: 'default' },
      { children: 'Valeriia Maltseva', iconName: 'user-01', color: 'default' },
      { children: 'Martin Alexander Feldkirchner', iconName: 'user-01', color: 'warning' }
    ],
    [
      { children: 'John Doe', icon: 'user-01', bordered: false },
      { children: 'Jane Doe', iconName: 'user-01', color: 'default' }
    ]
  ]
  }
/>
```


assets/js/src/core/components/tag-list/tag-list.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+interface StylesProps {
+  itemRowGap?: number
+  itemColGap?: number
+}
+
+export const useStyles = createStyles(({ css, token }, props: StylesProps) => {
+  return {
+    tagListGroup: css`
+      row-gap: ${props.itemRowGap ?? token.Tag.rowGapXXS}px;
+      column-gap: ${props.itemColGap ?? token.Tag.colGapXXS}px;
```
vin0401: I would prefer when we align this with our general spacing values like we have it in the Flex or Space Component.

We also could think of making a modification on the flex component.
So that we have the possibility to only use a normal string value, like we already have it right now.
And in addition you also could pass something like that:

``` TypeScript
<Flex gap={{ x: 'small', y: 'mini' }}>
    ...list of tags.
</Flex>
```

What do you think?

assets/js/src/core/modules/app/theme/utils/themes/default-theme.ts
```diff
@@ -182,11 +182,18 @@ const defaultTheme = {
...
       fontSize: 12,
       fontSizeIcon: 12,
       fontSizeSM: 12,
       defaultBg: '#fafafa',
       defaultColor: 'rgba(0, 0, 0, 0.88)'
+    },
+
+    Tooltip: {
+      colorBgSpotlight: 'rgba(0, 0, 0, 0.85)',
+      borderRadiusLG: 6
```
vin0401: Couldn't find these in the Theme Buddy. Are they aligned with @idaiv

vin0401 (changes requested):
I think we are already pretty close to the final solution. :P

assets/js/src/core/components/flex/flex.tsx
```diff
@@ -11,50 +11,62 @@
...
+      }
+    }
+
+    return { rowGap: 0, colGap: 0 }
   }
 
   return (
     <AntFlex
-      gap={ internalGap }
+      className={ flexClassNames }
```
vin0401, line 65: Does this still work, when I add a className in the props?

I think it would also be good to have an Example for the the new gap value. :)

assets/js/src/core/components/flex/flex.tsx
```diff
@@ -11,50 +11,62 @@
...
+
+    if (isString(gap)) return { rowGap: getGapValue(gap), colGap: getGapValue(gap) }
+
+    if (isNumber(gap)) return { rowGap: gap as number, colGap: gap as number }
+
+    if (isObject(gap)) {
+      return {
+        rowGap: getGapValue((gap as GapRowColGroupType).rowGap),
+        colGap: getGapValue((gap as GapRowColGroupType).colGap)
+      }
```
vin0401: Can we call it x and y like in my initial proposal? :)
I would be more we also use x and y for the padding. So it would be more consistent, when you use the different spacing components. What do you think?

assets/js/src/core/modules/app/theme/utils/themes/default-theme.ts
```diff
@@ -43,6 +43,7 @@ const defaultTheme = {
     colorFill: 'rgba(215, 199, 236, 0.6)',
     colorFillQuaternary: 'rgba(215, 199, 236, 0.4)',
     colorBgLayout: '#fcfcfc',
+    colorBgSpotlight: 'rgba(0, 0, 0, 0.85)',
```
vin0401: Can't see this token in theme buddy. Is it aligned with @idaiv?

assets/js/src/core/modules/app/theme/utils/themes/default-theme.ts
```diff
@@ -182,6 +183,8 @@ const defaultTheme = {
       lineWidth: 1,
       marginXS: 8,
       paddingXXS: 4,
+      colGapXXS: 4,
+      rowGapXXS: 4,
```
vin0401: Do we still need them? I think they are not in use? Also can't see them in Theme buddy

assets/js/src/core/components/tag/tag.tsx
```diff
@@ -38,19 +51,38 @@ export const Tag = ({ children, icon, iconName, theme, className, ...props }: Ta
...
   )
+
+  return (
+    <>
+      {shouldShowTooltip
+        ? (
+          <Tooltip
+            align={ { offset: [20, -4] } }
+            overlayClassName={ styles.tooltip }
+            title={ children }
```
vin0401: I'm not sure if the title should always be the direct children. I also could imagine that there is a longer descriptive text in there, compared to the content. Maybe we add a new Prop tooltip title? 

We also could omit the shouldShowTooltip prop then, and only show the tooltip when a tooltip title is passed as a prop.

vin0401: hmm, but that would also mean, that you can't really control the tooltip based on the maxLength anymore.
I don't know for me it seems like we are limiting us here a bit. Maybe it would be useful to have some kind of truncate utility and just handle the cutting of the content outside of the tag component? 

What do you think?

vin0401 (changes requested):

assets/js/src/core/components/flex/flex.tsx
ValeriaMaltseva: Thanks for the comment. I've fixed it.

assets/js/src/core/modules/app/theme/utils/themes/default-theme.ts
ValeriaMaltseva: I've deleted it, thanks.

assets/js/src/core/components/tag/tag.tsx
ValeriaMaltseva: > I'm not sure if the title should always be the direct children. I also could imagine that there is a longer descriptive text in there, compared to the content. Maybe we add a new Prop tooltip title?
> 
> We also could omit the shouldShowTooltip prop then, and only show the tooltip when a tooltip title is passed as a prop.

I see what you mean. However, on the other hand, it could be a bit tiring to pass _tooltipTitle_ each time. That’s why I propose to keep the current behavior as the default and additionally add _tooltipTitle_ prop to cover the case you mentioned earlier. What is your opinion?

ValeriaMaltseva: > hmm, but that would also mean, that you can't really control the tooltip based on the maxLength anymore. I don't know for me it seems like we are limiting us here a bit. Maybe it would be useful to have some kind of truncate utility and just handle the cutting of the content outside of the tag component?
> 
> What do you think?

From my perspective, it’s good to have the opportunity to set a character limit and show the Tooltip depending on the Tag content (we can either set this value or not). Otherwise, we would need to define rules for when to truncate the Tag content, which could lead to additional edge cases that we would need to handle.

ValeriaMaltseva: Lastly, showing a tooltip based on the maximum character length was a requirement from @idaiv . If we want to change it, we need to have an additional grooming session with Ida.

vin0401 (approved):

PR #569, opened 2024-10-04
Prepare context menu for main widget tabs

markus-moser:
image


assets/js/src/core/modules/widget-manager/hooks/use-context-menu.tsx
```diff
@@ -0,0 +1,91 @@
...
+      setContextMenuState({ x: event.clientX, y: event.clientY, tabNode: node })
+      node.getExtraData()
+    }
+  }
+
+  const closeContextMenu = (): void => {
+    setContextMenuState(null)
+  }
+
+  const handleClickOutside = (event: MouseEvent): void => {
```
vin0401: @Corepex defined a hook for this, some time ago:
https://github.com/pimcore/studio-ui-bundle/blob/1.x/assets/js/src/core/utils/hooks/use-click-outside.ts

Would it make sense to use it here?

vin0401 (changes requested):
Nice. :) Just one small thing. :)

assets/js/src/core/modules/widget-manager/hooks/use-context-menu.tsx
markus-moser: Ahh, cool thanks - looks like if it is exactly how we need it

markus-moser: done

vin0401 (approved):

Issue #510, opened 2024-09-02
Select component / styling


markus-moser:
Currently e.g. in the properties tab we have a special styling for select boxes:

https://github.com/pimcore/studio-ui-bundle/blob/1caa5f47dd4b8f09d3f1767588ec94bc44749fe0/assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/properties/properties-container.styles.tsx#L72

This transforms it from this

![image](https://github.com/user-attachments/assets/5aabcb90-0f19-46f8-9fbf-cc531878fcd6)

to this

![image](https://github.com/user-attachments/assets/49a269b4-7667-49bb-a508-67f139416b09)

We should provide a centralized select component which provides the styling options we need and re-use it everywhere.

Also in the data grid design we need to think about how select boxes should look like. The design of the scheduler e.g. looks like this:

![image](https://github.com/user-attachments/assets/803b724c-1943-440c-a878-552b7b0f6ab3)

I unified it to the default ant design select design. We should re-style it in the `SelectCell` centrally so that it looks the same everywhere.

ValeriaMaltseva:
**The design part:**
https://github.com/pimcore/product-management/issues/642#issue-2564741769
https://www.figma.com/design/lomefPHRhJnkljB2nXtuKV/Ant-Design-5.11?node-id=1226-85112&node-type=frame&m=dev

PR #581, opened 2024-10-09
[510] Create Select Component

ValeriaMaltseva:
**Task**: https://github.com/pimcore/studio-ui-bundle/issues/510


assets/js/src/core/components/select/select.styles.ts
```diff
@@ -0,0 +1,195 @@
+/**
```
vin0401, line 1: We use a lot of !important overwrites here? 
Do we really need them to this extend?

assets/js/src/core/modules/asset/editor/shared-tab-manager/tabs/custom-metadata/custom-metadata-container.tsx
```diff
@@ -150,6 +148,7 @@ export const CustomMetadataTabContainer = (): React.JSX.Element => {
                 />
 
                 <Select
+                  className={ styles.customMetadataSelect }
```
vin0401: Would it also be possible to use our utility classes for this case like "w-full"`?
That way we would be able to delete the related style file completely. :)

assets/js/src/core/modules/asset/editor/types/image/tab-manager/tabs/preview/sidebar/tabs/details/helpers/config.tsx
```diff
@@ -0,0 +1,88 @@
+/**
```
vin0401: I would prefer to see this file in a hooks directory and the file name should also reflect the hook name. 
We already have them here and there and normally we follow the same approach for all of them. :)

assets/js/src/core/modules/element/editor/shared-tab-manager/tabs/properties/properties-container.styles.tsx
```diff
@@ -13,27 +13,14 @@
```
vin0401: Same here, can we use our utility classes in that case? 
Because then we can delete this file completely

vin0401 (changes requested):
Looks already pretty good to me. But still, there are a few smaller improvements I would love to see, if possible.

assets/js/src/core/components/select/select.styles.ts
ValeriaMaltseva: I have already removed some of them, but in general, we should keep them because, on the AntD side, there are many selectors with higher specificity.

vin0401: Okay, should be okay I guess. My only concern is, that when we want to wrap the select in another component, with new styling it would be a bit harder to overwrite them. But the chances for that are quite low, I guess. So I think we are fine. :)

vin0401 (approved):

PR #631, opened 2024-10-21
Horizontal scroll

### LINKED ISSUE #576: Implement workflow transitions and workflow state next to save button

Related to #226

xIrusux:
Created Horizontal, tab-able scroll component. 
Scrolling can also happen by clicking the 'left' and 'right' icons or finger drag 

Made some small adjustments to the existing TagList Container 

Embedded new Scrolling Container in workflow menu

resolves [#576](https://github.com/pimcore/studio-ui-bundle/issues/576)

## Additional info

https://github.com/user-attachments/assets/a02a99a7-47a4-41ee-85f6-80bf9e7a1639


xIrusux:
Only show scroll width when required

xIrusux:
Resizing when there is no set scroll width: 

https://github.com/user-attachments/assets/2f40224e-b3c0-40c0-9e80-d2b7380a0b9b

Resizing when there is a set scroll width: 

https://github.com/user-attachments/assets/5c66fec4-8090-406f-8bb8-ed1972f32dc3

xIrusux:
https://github.com/user-attachments/assets/2d6ededd-5809-4a58-809a-f718ddcfbce2

Added illustration of what the resizing looks like to storybook. 
However in storybook it seems a tiny bit more jumpy than it is when embedded in studio ui

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+*  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import { createStyles } from 'antd-style'
+
+export const useStyles = createStyles(({ css }, { scrollWidth, hideElement }) => {
+  return {
+    scrollContainer: css`
+        visibility: ${hideElement === true ? 'hidden' : 'visible'};
```
vin0401, line 19: This line confuses me a bit. Would that not hide the complete element?

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
```diff
@@ -0,0 +1,154 @@
...
+          onMouseUp={ stopScrolling }
+          theme="secondary"
+        />
+      )}
+      <Flex
+        align={ 'center' }
+        className={ styles.scroll }
+        ref={ scrollContainerRef }
+      >
+        {hideElement ? children : children}
```
vin0401: When it's either children or children. Couldn't we use children here without the condition?

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
```diff
@@ -0,0 +1,154 @@
...
+      updateScrollState()
+      scrollContainerRef.current.addEventListener('scroll', updateScrollState)
+
+      const resizeObserver = new ResizeObserver(() => {
+        updateScrollState()
+      })
+
+      resizeObserver.observe(scrollContainerRef.current)
+
+      return () => {
```
vin0401, line 57: I think we should also clear the scrollInterval here. In case there is a weired bug this interval would never be cleared.

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const startScrollingRight = (): void => {
+    if (scrollInterval === null) {
+      const interval = setInterval(() => {
+        if (scrollContainerRef.current !== null) {
+          scrollContainerRef.current.scrollBy({ left: scrollAmount, behavior: 'smooth' })
+        }
+      }, scrollSpeed)
+      setScrollInterval(interval)
+    }
+  }
```
vin0401: We can remove some duplications here, right?
Having a base method that accepts either the negative or the positive scrollAmount could handle both cases, right?

vin0401 (changes requested):
Like the approach. :) But would love to see some small improvements. :P

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.styles.ts
xIrusux: yes here it is completely hidden which makes sense when there is very little space. Otherwise one just sees the scroll right and scroll left button without anything inside it

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
xIrusux: of course - this was left over I missed.
I just updated it.

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
xIrusux: agree - and have just added it

assets/js/src/core/components/horizontal-scroll/horizontal-scroll.tsx
xIrusux: done :)

xIrusux:
Screenshot 2024-10-24 at 11 49 49

There is still a problem when we have a very very very small width available with the button becoming scrollable. 
I think it would be good if we would implement the "visibility" hidden more widely to work with such situations. 

Let me know what you think!

vin0401 (approved):

PR #656, opened 2024-10-28
Full width horizontal scroll

### LINKED ISSUE #576: Implement workflow transitions and workflow state next to save button

xIrusux:
resolves https://github.com/pimcore/studio-ui-bundle/issues/576

- Rework of the horizontal scroll bar to remove the use of global overflow-x: auto class which was required when the bar is nested in several flex boxes
- Therefore encapsulate all required behaviour in the shared component itself. 
- Change of focus to have the scroll bar span across the entire width of the toolbar which can be used in other places as well. 
- Initially tried to have an overall horizontal scroll bar for the entire width of the toolbar and inside that another horizontal scrollbar. However this brings back the requirement of adding overflow-x: auto back or passing args through both nested horizontal scroll bars to determine which needs to shrink first and when etc.

## Additional info

https://github.com/user-attachments/assets/4ce4f078-a991-4ac4-83ce-0ec9e0abd96e


xIrusux:
implemented at a different level so the scroll automatically appears for the toolbars + pagination toolbars 

https://github.com/user-attachments/assets/20d98903-85a6-4711-96c1-f14e5902a2d0

vin0401:
Closed due to https://github.com/pimcore/studio-ui-bundle/pull/661

PR #663, opened 2024-10-31
Workflow actions

### LINKED ISSUE #576: Implement workflow transitions and workflow state next to save button

xIrusux:
Resolves [#576](https://github.com/pimcore/studio-ui-bundle/issues/576)

## Additional info

https://github.com/user-attachments/assets/3914d7ae-dfe9-4bb6-8f1a-86524e9d58cc


xIrusux:
https://github.com/user-attachments/assets/2c0af7ea-d3c6-451d-b5c7-ec2a3a8f5882

@vin0401 since we last spoke :) 

- added an error snackbar
- used redux shared mutation results -> https://redux-toolkit.js.org/rtk-query/usage/mutations#shared-mutation-results
which mean button are switched to loading state when submitting an action. However we can still not distinguish between which button should be in loading state and which should not due to us not sharing the "workflowDetails" Context between different instances of the Provider (some follow up solution still needs to be found here)
- fixed up pop up based on Ida's feedback
- fixed up small ux issue where there was no longer a gap in the toolbar when the horizontal scroll does not exist 
- ensured we use translations where possible 

I have started a convo re how it should look when the workflow cards are loading after submission an action

xIrusux:
Tried many things to create a good way of loading inside the sub groups in the toolbar dropdown. 
This one works but still has issues with the backgrounded button while loading. 
Take a look below

https://github.com/user-attachments/assets/61048777-3cdb-4cf2-a3a7-3f1fa03252be

vin0401 (approved):

xIrusux:
Thanks for the approval :) 

Final changes were: 
- create custom "workflow transitions group" that bundles several dropdown button into one custom item 
- therefore the separate dropdown items are styled in one file per workflow and can be replaced by one loading button which really improves the feel over the last iteration where each workflow transition button had a loading state i.e. 10 workflow transitions -> 10 loading buttons.
- the button has a side effect when inside the ant d dropdown which has to do with the animated custom loading animation that is inside. This is a very fiddly topic that has to be looked at separately. 
- code to show workflows as loading per workflow submission has been removed. This is because we decided it is not good enough to show it as only reloading the workflow that has a submitted transition when the reality is that all workflows get reloaded

Issue #176, opened 2024-03-28
Asset List Search & Filter


fashxp:
![image](https://github.com/pimcore/product-management/assets/8792145/598422ea-55a1-4805-8a69-54b28badd574)

ValeriaMaltseva:
The front-end part consists of four parts and will be implemented in four separate PRs.

1. ✅ Full-text search filter (the back-end part will be implemented in this task: https://github.com/pimcore/studio-backend-bundle/issues/523).
2. ✅ "only direct children" checkbox filter (the "only unreferenced" is currently on hold).
3. ✅ PQL filter.
4. Field filters: 
     1. ✅ Fix ID field.
     2. ✅ Add Datetime field.
     3. ✅ Add Fullpath field.

PR #669, opened 2024-11-04
Add PQL filter

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/176


assets/js/src/core/modules/asset/editor/types/folder/tab-manager/tabs/list/list-provider.tsx
```diff
@@ -225,26 +226,34 @@ export const ListSortingProvider = ({ children }: ListSortingProviderProps): Rea
 }
 
 export interface IListDataContext {
-  data: AssetGetGridApiResponse | undefined
+  data?: AssetGetGridApiResponse
+  errorData?: FetchBaseQueryError
```
vin0401: hmm, not sure if we should place the error information that are filter related in here.
The data context is holding the data that is shown in the grid.

Why are you putting this here instead of the filters?
What do you think about moving this to the filters?

assets/js/src/core/modules/asset/editor/types/folder/tab-manager/tabs/list/sidebar/filters/filter-container-inner.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+* Full copyright and license information is available in
+* LICENSE.md which is distributed with this source code.
+*
+*  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import { createStyles } from 'antd-style'
+
+export const useStyles = createStyles(({ css, token }) => {
```
vin0401: Can we move these styles to our UI components somehow?
I guess it's not the last time we encounter these stylings? :)

assets/js/src/core/modules/asset/editor/types/folder/tab-manager/tabs/list/sidebar/filters/filter-container-inner.tsx
```diff
@@ -72,39 +106,101 @@ export const FilterContainerInner = (): React.JSX.Element => {
...
-            {/* </Checkbox> */}
-          </Space>
-        </Form>
-
-        <Title>
-          Field filters
-        </Title>
-
-        <FieldFiltersContainer />
+              <Flex gap='mini'>
```
vin0401: Can we create a stand alone component for this PQL Filter/Input? 
At a later state of the project, we will use this component in several places and bundles. 

And also the idea is to integrate some syntax highlighting in the future. 
What would also imply to move this to a separate component. :)

vin0401 (changes requested):
Love it already. :) 
But I marked a few things, where we should have a second look. ;)

assets/js/src/core/modules/asset/editor/types/folder/tab-manager/tabs/list/list-provider.tsx
vin0401: Or to be more precise to move this to the filter context. :P

ValeriaMaltseva: I was thinking about this and decided to proceed with this simplified logic as otherwise we would need to update each provider that uses specific filters. However, I agree that the alternative approach is ideologically better, so I've updated the logic accordingly. Thanks for highlighting it! :)

vin0401 (approved):

PR #677, opened 2024-11-07
Add simple alert modal

### LINKED ISSUE #628: Add alert modal
Extend the prompt + confirm modal functionality by alert messages.

We will mainly use it for error messages like this example from the classic bundle:
Image

(we can skip the "details" button for now)

Idially it would be possible to have two different versions for `info` and `error` with different icons or slightly differnt stylings.

Design: https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=157-58069&node-type=canvas&t=st4QjUicCjKfFtbC-0

Corepex:
Resolves #628


assets/js/src/core/components/modal/simple-modal/hooks/use-simple-modal.tsx
```diff
@@ -0,0 +1,59 @@
...
+  content: string
+}
+
+export interface UseSimpleModalResponse {
+  info: (props: ContentAware) => { destroy: () => void, update: (configUpdate: ConfigUpdate) => void }
+  error: (props: ContentAware) => { destroy: () => void, update: (configUpdate: ConfigUpdate) => void }
+  warn: (props: ContentAware) => { destroy: () => void, update: (configUpdate: ConfigUpdate) => void }
+}
+
+export const useSimpleModal = (): UseSimpleModalResponse => {
```
markus-moser: Could you please rename this to `useAlertModal()`. I thought you will integrate this to the existing `useFormModal()` - this is why I suggested `useSimpleModal` as name. For me it's fine when this is a separate helper, but then `useAlertModal()` is a better name.

markus-moser (changes requested):
Other then that it LGTM

assets/js/src/core/components/modal/simple-modal/hooks/use-simple-modal.tsx
Corepex: Okay, I misunderstood that. I thought this would be the way. Nevertheless, I would go with this solution since it's cleaner.
I suggest reworking the original model to the new (and much smoother) approach.

markus-moser: Yes, if you find the time feel free to rework the useModal() hook to the new approach 👍

markus-moser (approved):

PR #699, opened 2024-11-18
Workflow objects

### LINKED ISSUE #698: Data object workflows

xIrusux:
Resolves #698

## Additional info

- show workflow toolbar in objects 
- distinguish between workflow transition submission api call asset / object
- issue outstanding with backend [#555 ](https://github.com/pimcore/studio-backend-bundle/issues/555) (to handle api call correctly)
- hide dropdown if no workflows available
- the backend issue pointed out some problems in the existing error handling -> I thought RTK query takes care of HTTP errors -> however a specific handling was required here and i learned about unwrap + fixed this


assets/js/src/core/modules/asset/editor/toolbar/workflow-log-modal/hooks/use-workflow.ts
```diff
@@ -32,9 +33,20 @@ export type TransitionType = 'transition' | 'global'
...
+
+  const assetHasWorkflowAvailable = asset?.hasWorkflowAvailable ?? false
+  const objectHasWorkflowAvailable = dataObject?.hasWorkflowAvailable ?? false
+
+  const hasWorkflowAvailable =
+      elementType === 'asset'
+        ? assetHasWorkflowAvailable
+        : elementType === 'data-object'
+          ? objectHasWorkflowAvailable
+          : false
```
vin0401: Can't you use the element draft in that case? Because the workflows will also be available for documents. 
And it would remove the need for all these checks. :)

vin0401 (changes requested):
There is just a small improvement that I would like to see. :)

assets/js/src/core/modules/asset/editor/toolbar/workflow-log-modal/hooks/use-workflow.ts
xIrusux: ah super cool! - I had not known about that option. have changed that now :)

vin0401 (approved):

Issue #719, opened 2024-11-21
[Asset List Filters] Switch behaviour of "only direct children" checkbox


markus-moser:
![Image](https://github.com/user-attachments/assets/fe6ea095-3898-43de-8f3a-eaa3875036ab)

* When the checkbox is not set "includeDescendants" needs to be true in the API
* Currently this lead also to the situation that when I open the list view of a folder without filtering, that the folder is empty (if it does not have any direct children):
![Image](https://github.com/user-attachments/assets/f93c3f89-443d-4a1a-8354-b1621bbaf7cc)

PR #726, opened 2024-11-21
Fix the logic of the includeDescendants filter

ValeriaMaltseva:
Tasks:
- https://github.com/pimcore/studio-ui-bundle/issues/176
- https://github.com/pimcore/studio-ui-bundle/issues/719


markus-moser:
@ValeriaMaltseva Thanks for the quick fix - I gave it a try and with your PR it now works as expected

vin0401 (approved):

PR #743, opened 2024-11-26
Batch edit event bus improvement

### LINKED ISSUE #648: [Bug] batch edit does not work
**Steps to reproduce**

- Login as admin
- Open a folder 
- Switch to the list view
- Mark some assets
- Start a batch editing process
- Edit 'Title' and 'Author'

**Actual behaviour**
Image
Image

- Batch edit did not work

**Expected behaviour**

- Values should be applied to the marked assets

**Environment**

- Nightly build

xIrusux:
Resolves #648

## Additional info
initially worked on the promise handling to ensure the reload happens at right time but the solution eventually was to
change when we publish the refresh event to the event bus to only happen on "finished" status of the job. This avoids reloading too early


vin0401 (approved):
Looks good to me. :)

Issue #639, opened 2024-10-24
Error handling for ajax requests


markus-moser:
Analyze ajax requests, find reasonable approaches to deal with errors and implement them where not already done.

PR #752, opened 2024-11-28
Error Handling

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/639


assets/webpack.config.js
```diff
@@ -185,6 +185,13 @@ let config = Encore.getWebpackConfig()
 config = {
   ...config,
 
+  devServer: {
+    client: {
+      overlay: false,
+      logging: 'none'
```
vin0401: We shouldn't disable that completely. It would be good, when webpack would still report the compilation errors. 
And since we are only catching API Erorrs so far, I'm also not sure about the runtime errors.

assets/js/src/core/modules/app/app-loader.tsx
```diff
@@ -31,17 +34,26 @@ export interface IAppLoaderProps {
 
 export const AppLoader = (props: IAppLoaderProps): React.JSX.Element => {
   const dispatch = useAppDispatch()
-  const [translations] = useTranslationGetCollectionMutation()
   const { i18n } = useTranslation()
+
   const [isLoading, setIsLoading] = useState(true)
+  const [errorData, setErrorData] = useState<FetchBaseQueryError | SerializedError | null | undefined>(null)
```
vin0401: I think this is still a bit more work than needed.
That would be the recommendation from my side. 

``` typescript
class APIError extends Error {
  // .... Extend it until it can handle the Errors from RTK-Query
}

const MyComponent = () => {
  const { report } = useErrorHandler();
  const { isLoading, isError, error} = useApiHook() // <= some GET request example
  const [ 
    fetchMutation, 
    {  isError, error } // only use these when you want to force a rendering error, so that it get catched by error boundary
  ] // some Post request / mutation example

  // Should be catched by error boundary. Because it's a error during the render of the react app
  if (isError) {
    report(new APIError(error))
  }
  
  const onClick = () => {
    fetchMutation()
      .catch(() => {
        // need to be catched by ErrorHandler since the error boundary can't detect that error, because this is not causing rendering issues
        report(new APIError(error))
      })
  }
  
  return (
    <Button onClick={onClick}>My button</Button>
  )
}
```

assets/js/src/core/hooks/use-api-error-handler.tsx
```diff
@@ -0,0 +1,58 @@
...
+type UseApiErrorHandler = (errorData: FetchBaseQueryError | SerializedError | undefined | null) => void
+
+interface IErrorData {
+  detail: string
+  message: string
+}
+
+const DEFAULT_ERROR_CONTENT = 'Something went wrong.'
+
+export const useApiErrorHandler: UseApiErrorHandler = (errorData) => {
```
vin0401: I think we should design this a bit more generic. 
The odds are not to small that we want to report other Errors, than just APIErrors in the future. 

Like mentioned in the code sample I would report real errors. 
And based on the type `instanceof error === 'APIError'` we can handle the different cases.
That would even cover inheritance and would allow us to even define different behaviors for different error types.

Or we could even open this further up, to also allow the community to define there own error types and handling.

assets/js/src/core/modules/app/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (!isEmpty(fallback)) return fallback
+
+      return (
+        <Flex
+          align='center'
+          gap={ 10 }
+          justify='center'
+          style={ { position: 'absolute', inset: 0 } }
+        >
+          <Typography>{error?.message ?? 'Something went wrong.'}</Typography>
```
vin0401, line 60: We could think of a try again button.
That renders the children again, in case there was just a temporary error.

When we catch the errors from my code example here we could also think of giving these back to our error handler, to show some more detailed errors as alerts. :)

vin0401 (changes requested):
@markus-moser @ValeriaMaltseva 

What do you think about my comments? 
I still would love to see a bit different approach here. 
Just let me know, when I'm on a completely wrong track. xD

assets/js/src/core/modules/app/error-boundary/error-boundary.tsx
markus-moser: Yes for some cases a retry button would make sense, but I would not do this in the first version and also make it dependent on the error type.

For the additional error information I totally agree - with different types of errors we could also handle different levels of error details.

assets/webpack.config.js
markus-moser: We should definitely not disable all logging.

assets/js/src/core/hooks/use-api-error-handler.tsx
markus-moser: Yes we should have different types of errors and all errors should be catched centrally in the error boundary and based on the type of error the boundary should decide how to handle/display them. But one use hook specialized for the api errors like this could in my opinion still make sense, wouldn't it? At the end this hook could also trigger/throw the regular error types but it would just be an helper to create the right (default) error type with the right (default) content.

assets/js/src/core/modules/app/app-loader.tsx
markus-moser: I agree on that approach in general but instead of `report(new APIError(error))` a small helper which does this based on the response would be an advantage in my opinion.

assets/js/src/core/hooks/use-api-error-handler.tsx
ValeriaMaltseva: @vin0401
Could you elaborate in more detail, please? What other errors do you mean?

From my perspective, it still seems beneficial to have a separate hook dedicated only to API errors, with the flexibility to update it for any new requirements in the future. In my opinion, the main drawback of having common hook for all types of errors is the potential for type conflicts. Given that the API is a significant part of the project, this approach seems to introduce unnecessary complexity.

Currently, I see two options: handling errors from the API separately and catching all others in the ErrorBoundary. To me, this approach seems to be enough.

What is your opinion?

assets/js/src/core/modules/app/error-boundary/error-boundary.tsx
ValeriaMaltseva: @vin0401 @markus-moser 

I was thinking about the "try again/retry" button, whether in the context of ErrorBoundary or API error handling. In my opinion, it makes more sense for API errors (for example, to retry a request, which is particularly useful for handling network connection issues). I also agree with Markus that it would be better to add this functionality later, not in first version or when we have specific requirements for it.

vin0401: yep, we can postpone that. 

But we should keep in mind that the error boundary is only catching exceptions that are preventing a part of the app from rendering and will render a fallback component instead. The error boundary will not catch exceptions that occur in a click Event for example. 

That means in the end that we need it for both the error boundary as well as for the the error handler.

assets/js/src/core/hooks/use-api-error-handler.tsx
vin0401: In general I'm fine with having a separate handler for the APIErrors. 
But these are not the only errors that can happen. 

For example we could have a valid response from the API but there is a piece of information missing - maybe in some kind of configuration that prevents us from processing whatever we are trying to do at that moment. For sure there are also other cases. And for that it would be good to have a general error handler in hand that allows us to handle these errors. Or even better we would have the possibility to describe how a error should be handled by this global error handler. 

This is just a rough idea so don't take it to serios just want to point out what I have in mind. :D 

``` typescript
abstract class AbstractErrorHandler {
    processibleErrors: (typeOf Error)[]
    handleError(error) {}
    // ...
} 

class APIErrorHandler extends AbstactErrorHandler {
  processibleErrors: [APIError]
  
  handleError() {
    // describe handling of API Errors
  }
}

class ErrorProcessor {
  errorHandlers: AbstractErrorHanlder[]
  
  // register a new errorHandler
  register(errorHanlder) {}
  
  // check for a matching error handler and process the error
  process(error) {
    const errorHandler = errorHanlders.find(handler => handler.processibleErrors.includes(instanceOf error))
    
    errorHandler.handleError(error);
  }
}

// ----
// somewhere at the early stage of the Application
errorProcessor.register(new APIErrorHanlder());
// ----

const useErrorHandler = () => {   
   report(error) {
     throw new error;
  }
}

class ErrorBoundary extends React.Component {
  // catch the error via react Error Boundary or via window.on('error', {}) depending on the origin of the error either react rendering issue or async event. And process it via ErrorProcessor.process(error)
}
```

That way we would have the possibility to describe the handling for all kind of different Errors and on the other hand could also handle unknown ones or just rethrow them. We could even think of adding a priority in the future, in case we have ErrorHandlers that handle the same ErrorTypes.

assets/js/src/core/modules/app/app-loader.tsx
ValeriaMaltseva: @vin0401 I like your approach, already updated it based on your suggestion ☺️

assets/js/src/core/modules/app/error-boundary/error-boundary.tsx
ValeriaMaltseva: @vin0401 Agree with you. Think, we need to consider adding try/catch logic for them when we need to track it for some reason.

assets/js/src/core/hooks/use-api-error-handler.tsx
ValeriaMaltseva: Thank you for so detail explaining, I catch your thoughts and on general like it :) However, I still have concern that there may not be many cases for the type of errors you described as examples and that we will have the same behavior how to handle it. For this reason, I'm not sure it makes sense to expand the logic as I would prefer to implement global logic only when there is a common need, such as handling API errors.

To bottom line, I would like to suggest in current implementation go with the logic of having only API error handling. However, if we see in the future that we need to handle other types of errors and this handling should be consistent, let's expand the approach.

@vin0401 @markus-moser What are your opinion?

vin0401: We already have a set of errors which are already thrown:
https://github.com/search?q=repo%3Apimcore%2Fstudio-ui-bundle+throw&type=code&p=1

These also need to be shown to the user. 
Most of them will end up at the error boundary, but still we want to control the behavior of these Errors in the the boundary and the errorHandler would be the best approach for that in my opinion. 

So having this in place would be really useful and would prevent us on searching for custom solutions that all work pretty similar. 

I'm open for other approaches but I definitely see the need to unify the logic for these.

ValeriaMaltseva: @vin0401 @markus-moser 
I'll take in work the last point from Marvin, and it generally looks like we are on the right track. Thanks for sharing your ideas and thoughts, a real teamwork vibe 😍

markus-moser: I totally agree with the suggested ideas of @vin0401. As it is a really important topic we should invest the time to make it flexible to be able to work for various types of errors.

assets/js/src/core/modules/app/error-handler/error-handler.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ErrorModalService } from '@Pimcore/modules/app/error-handler/services/error-modal-service'
+
+interface IErrorContentProvider {
+  getContent: () => string
+}
+
+const trackError = (data: IErrorContentProvider): never => {
+  const errorContent = data.getContent()
+
+  ErrorModalService.showError(errorContent)
```
vin0401: Really like the overall approach now. :)

One last thing, when we can easily achieve it. 

Maybe we can inverse this dependency here?
So that we can register different services beforehand for the different error types?

That would make it super flexible and would allow us to define the handling for each error type in more detail, if needed.

vin0401 (approved):

Issue #730, opened 2024-11-21
[Tags] Other elements get reloaded when adding a tag


markus-moser:
* Open 2 assets in parallel
* Add/delete a tag
* Take a look at the chrome network tools. The tags are not only refreshed for the asset where you currently work on but also for the other one.

Issue #724, opened 2024-11-21
[Tags] Refresh tags on element reload


markus-moser:
Currently it seems that the RTK query tags are not setup/cleared correctly when an element gets reloaded (either via reload button or by closing the element and re-opening it).

The tags do not get refreshed from the server.

PR #792, opened 2024-12-03
Tags loading

### LINKED ISSUE #725: [Tags] Loading indicator
In the element tags tab we implemented optimistic update when adding/removing tags. We need a loading indicator while the requests get executed.

xIrusux:
Resolves #725 and #730 and #724

## Additional info
 
- Tags have been allocated so we are only fireing the required POST's / DELETE's when adding / removing tags 
- The Tags cache gets invalidated when the Tabs component gets re-rendered (i.e. when another user makes a change to the tags, we have to close the tab and reopen it to see the change ) 
- There is an error snackbar / message at the bottom of the screen if assigning / un assigning a tag does not work
- In case of error we also undo the optimistic loading 
- The Tag checkbox enters a loading state during POST/DELETE and further actions are not allowed during that time
- There is no more cross refreshing when other assets with the tags tab open are opened as per 730


vin0401 (approved):

xIrusux:
@markus-moser 
Since we last spoke I have simplified the rtk tag set up and understood a few things about it I previously did not :). 
I also found that the Problem with the indicator for clicking the trash bin is a bit more complex than I thought. This is something I am solving as part of a separate pull I am currently working on. 

This changed notably:
- The POST/ DELETE requests when clicking the Tag checkboxes now have a GET response at the end. This is only one response and ensures the optimistic update gets verified and therefore logically makes sense. 
- I also removed the getElementDetailTag from the Tags and Notes and Event reloading invalidation Tags. This is as we do not seem to require a reload of the Asset when adding or removing Notes. 

Please let me know if you have any questions or there is something we should discuss :).
Thanks

markus-moser:
@xIrusux Thanks for the update, looks much better. Unfortunately I still sometimes have a strange behaviour with the checkbox states. Not sure if this is somehow connected with the `GET` request which refreshes it at the end. Please take a look at the video (I throttled down the network speed a bit for the video):


https://github.com/user-attachments/assets/2ffff8d1-8bc3-4651-9cfc-2bd35f911eb9

xIrusux:
Taking another look at this.
Thanks for testing!

xIrusux:
@markus-moser 

- tag deletion does now work and not fall over itself 
- the results are ordered alphabetically and one can choose the desc / asc in column header.  (There are slight differences in how the Tag tree on the left and the grid on the right are handling upper cases as children item paths i.e. ("Color/Red" vs "Color/red") In the Tag tree this is always put at the first place. While in the Grid the upper case is not always treated the same way. 


https://github.com/user-attachments/assets/911eca33-2d32-4820-8f00-4a7898d0a434

- In a separate issue I am exploring a separate ordering function and also a solution to the "Schedules" Tab not updating.

- I tested the other Tabs and did not find the same refresh issue except for "Schedules"

- When the assign or unassign fails - the optimistic loading reverts back to display the previous state

PR #808, opened 2024-12-09
Image gallery sorting drag and drop

markus-moser:


assets/js/src/core/components/drag-and-drop/context-provider.tsx
```diff
@@ -66,6 +68,8 @@ export const DragAndDropContextProvider = ({ children }: { children: ReactNode }
   function onDragStart (event: DragStartEvent): void {
     const data = event.active.data.current as DragAndDropInfo
     setContext(data)
+
+    console.log('start', event)
```
vin0401: I think we can remove the comment and the logging?

vin0401 (changes requested):
Just a bit of clean up, but looks good to me. :)

markus-moser:
Still room for slight styling improvements, but let's keep it like this for now.

PR #813, opened 2024-12-09
Use actual image dimensions for preview image thumbnails

markus-moser:
Needs some investigation on backend side before we can merge it:
https://github.com/pimcore/studio-backend-bundle/issues/624


markus-moser:
Already solved with a different PR

Issue #740, opened 2024-11-26
[Asset Image] Implement edit tab


markus-moser:
Use the same approach/library like in the classic bundle:

![Image](https://github.com/user-attachments/assets/64659101-2989-4592-be3c-2b89b3e67a94)

markus-moser:
* Integrate minipaint (or alternatively check if we find an even better tool as replacement) into the edit tab of image assets

* Take a look at this PR to get an idea how minipaint was integrated in the old UI:
https://github.com/pimcore/admin-ui-classic-bundle/pull/537

* The integration is mostly done in this file (HTML markup + JS):
https://github.com/pimcore/admin-ui-classic-bundle/blob/3deea54d4a78fffee4611d6f8f7edb985db8e82c/templates/admin/asset/image_editor.html.twig

* Ensure that we use the latest version of minipaint and try to integrate it into our webpack build. See https://github.com/viliusle/miniPaint/releases

* Check if we need additional endpoints from the backend team for saving/updating the image asset file content

markus-moser:
Minipaint is integrated as <iframe> in the old UI:

![Image](https://github.com/user-attachments/assets/27522d8b-544c-4270-91b0-79d2e365f6c9)

Would be ok if we also do it as iframe. Maybe we then could also handle it as twig template to simplify things. I can help you with the "backend" part of the iframe if we decide that we want to do a similar approach like in the classic bundle.

In that case we would not need to integrate it into webpack but could just use the packed/minified version of minipaint directly.

ValeriaMaltseva:
Blocked by this task: https://github.com/pimcore/studio-backend-bundle/issues/638 - RESOLVED

ValeriaMaltseva:
Blocked by this bug: https://github.com/pimcore/studio-backend-bundle/issues/668 - RESOLVED

PR #817, opened 2024-12-11
The MiniPaint integration

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/740


vin0401 (changes requested):
I saw that you introduced a new folder called services. 
I'm not sure what the purpose/meaning of this folder is.

Could elaborate a bit on that?
Right now, we already have a lib folder maybe that one would also be suitable not sure.

For me services are more like useful classes/objects that are provided via our dependency injection container. So I'm not sure if the name is really suitable in that case.

vin0401 (approved):

templates/image-editor/editor.html.twig
```diff
@@ -0,0 +1,221 @@
...
+
+    async function getTranslation() {
+        const url = '{{ path('pimcore_studio_api_translations') }}';
+
+        const response = await fetch(url, {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({ locale: 'en', keys: [] })
```
markus-moser: We should add the requested translation keys here in the keys array. With this approach we would only get the relevant translation keys instead of the full list.

markus-moser:
@ValeriaMaltseva Only found one small topic (see my comment). Other then that I tested it a bit and everything seems to be fine. We can merge it after checking my comment. 👍

PR #834, opened 2025-01-03
Icons update

### LINKED ISSUE #809: [Icons] Replace icons with versions from Figma and automate the process
* Export all icons from the Figma icon library into our file system
* Create a node/npm task to automatically generate our icon integration files
* Replace the icon names everywhere with the new names from Figma

xIrusux:
Resolves #809

## Additional info

To run the script go to studio-ui-bundle/assets and run:
**npm run generate-icons**

Find a quick "how to" at the top of the file:
image

__________

I have updated the tsconfig.json and feel like this may have an impact. 
image

Some naming changes should be made to the backend  IconService.php 
image

__________
**Details on Icons used / topics to discuss :** 

I have added icons / copied old ones and used icons that deviated to some extend from the old look to ensure all bases are covered. 
This needs to be design reviewed.  

Base Icons that were used:
https://www.figma.com/design/Q71sBiSRmHUDZhIqwEMVkm/Icon-Library?node-id=465-21&p=f&m=dev

1. reused what used to be the Icon "PlusCircleOutlined" under the name "new-circle"
image

2. used "file outlined" to replace the old "document" Icon

3. added the icons identically to the old icons: 
    - data-sheet
    - data-management
    - brush (used in cache)
    - minus

4. I have noticed that the icon named "details" seems out of place now. 
Below one can see how it has changed
image

5. there is a variant of data-object (one dark one light). Can we check which one should be used where? i.e. if everything is in the right place now.

6. the new icons also removed filled in circular warning, question mark etc. icons i.e. 
image

7. there are two edit pens in the current icon library which we might want to revisit

8. I used "content" to replace the old "note"
image

9. The below Item is called "group" which does not fit what is on the svg
image
It can now be found in the list overview and we should revisit the naming.


assets/package.json
```diff
@@ -17,7 +17,8 @@
     "lint": "eslint --ext .js,.jsx,.ts,.tsx ./js",
     "lint-fix": "eslint --ext .js,.jsx,.ts,.tsx ./js --fix",
     "test": "jest ./js/src --passWithNoTests",
-    "storybook": "storybook dev -p 6006 --no-open"
+    "storybook": "storybook dev -p 6006 --no-open",
+    "generate-icons": "ts-node -r tsconfig-paths/register js/src/core/modules/icon-library/generate-icon-index.ts"
```
markus-moser: Should we call ts-node with `npx ts-node` to avoid that it needs to be installed directly on the system?

I took a quick look at the options of ts-node and found two which should make it possible to overrule the compiler options from the defaut tsconfig.json:

`-O, --compilerOptions [opts]    JSON object to merge with compiler options`
`-P, --project [path]            Path to TypeScript JSON project file` 

Maybe the `--compilerOptions` is the simplest way to overrule the tsconfig and avoid that we need to change our default `tsconfig.json`?

Also I'm not sure if the location inside of modules is the best one for the `generate-icon-index.ts`. As it is a script related to the build maybe it's better to put it into `assets/build`?

xIrusux: did exactly this :) great suggestion

xIrusux:
solved smaller icon inconsistencies and now detects when a svg file does not have a "stroke" in this case "fill" is being used
fixed 9.

xIrusux:
cleaned up storybook.

xIrusux:
for the record. 
Tried to use the existing Icon and adjust via currentColor and a variable that is passed in.
Those variables are fiddly to adjust and its just way less intuitive than going with currentColor. 
When one uses an icon like below in the second row, there is only one color to be dealing with:

image

idaiv:
We have covered all the points on a call and most were resolved. On my side a couple of updates:
1. Clear cache: https://www.figma.com/design/Q71sBiSRmHUDZhIqwEMVkm/Icon-Library?node-id=817-2042&t=lMUUHgE6Fo0IsWan-4
2. Signs/ indicators for the modals - outline and filled icons: https://www.figma.com/design/Q71sBiSRmHUDZhIqwEMVkm/Icon-Library?node-id=637-1775&t=lMUUHgE6Fo0IsWan-4
3. Alternative to the details icon that you have mentioned feels off. We will apply the new icon only if the current choice is not approved. I really want to have less looking like report/file/document icons visually thats why I was vouching for this one being different: https://www.figma.com/design/Q71sBiSRmHUDZhIqwEMVkm/Icon-Library?node-id=832-992&t=lMUUHgE6Fo0IsWan-4

PR #849, opened 2025-01-09
Update README.md

dpfaffenbauer:
Restructure the plugin readme, I find it a bit confusing to find essential information at the end.

Also re the DllReference, please someone can comment on my comment? it does not work for me...


dpfaffenbauer:
also the other examples don't really work... I get the error:
```
Init my plugin
main.ts:11 Start up my plugin
core-dll.js:2 Uncaught (in promise) TypeError: e.onInit is not a function
    at core-dll.js:2:1638256
    at Array.forEach (<anonymous>)
    at e.value (core-dll.js:2:1638233)
    at core-dll.js:2:2113377
```

vin0401:
Hi @dpfaffenbauer 👋 

thanks for checking out the plugin system. :)
I agree with you - the npm installation should always comes first.

doc/05_Plugins/README.md
```diff
@@ -88,10 +103,12 @@ Encore
...
   // Important! Reference this vendor-manifest in your build.
   // It will take care of injecting Ant-Design, React, etc. without the need to bundle it in your plugin.
-  .addPlugin(new webpack.DllReferencePlugin({
+  // Note: This is currently not even possible, since the build dir changes with every build (uidv4) and even if you
+  // use the right folder, it fails... is it really needed?
+  /*.addPlugin(new webpack.DllReferencePlugin({
     context: __dirname,
     manifest: path.join(__dirname, 'node_modules', 'pimcore-studio-ui', 'dist', 'vendor',  'vendor-manifest.json')
-  }))
+  }))*/
```
vin0401: In general it's needed to ensure that always the same copy of react is used through the whole app. 
Otherwise react will start to complain latest at the point where you are using some of our react hooks in your plugin. Also it reduces the bundled size of the plugins since webpack is using a dynamic link to a shared module instead of bundling a new copy into every plugin.

The uuid folders shouldn't affect you. Have you referenced this file from the installed npm package of studio in your plugin? Can you share some more information what is exactly failing in this case?

In general webpack is mapping the imported modules over deterministic ids at runtime. So, even when the output files are placed somewhere completely different - webpack will be able to map modules due to these ids.

vin0401:
> also the other examples don't really work... I get the error:
> 
> ```
> Init my plugin
> main.ts:11 Start up my plugin
> core-dll.js:2 Uncaught (in promise) TypeError: e.onInit is not a function
>     at core-dll.js:2:1638256
>     at Array.forEach (<anonymous>)
>     at e.value (core-dll.js:2:1638233)
>     at core-dll.js:2:2113377
> ```

The onInit-Method in this case should be related to the plugin-system, I guess. 
But so far we can't really reproduce the error. Can you give us some additional information or can share some code with us, so that we can reproduce this error?

doc/05_Plugins/README.md
dpfaffenbauer: thanks for the response.

I figured it out now. I was confused due to the npm install of the studio ui. I though I needed to reference the actual build files in the public folder of the Studio UI Bundle, but I need to reference the node_modules studio ui bundle. with that, this is working now.

dpfaffenbauer:
@vin0401 here is what I did so far: https://github.com/dpfaffenbauer/CoreShop/commit/1f96fac24cf6ce19d37112433d3adde4e6529d44

vin0401:
@dpfaffenbauer - I was able to figure it out. :)

The error you send us is triggered due to this line:
https://github.com/dpfaffenbauer/CoreShop/commit/1f96fac24cf6ce19d37112433d3adde4e6529d44#diff-21fc732d543f3e01610a5a9dd7eba8b9a373cde2038f663acf4a40e5d9d71597R15

You are injecting a react component in this case but it should be the folderTabExtension-Module you created.
Also there are two cases where our docs are a bit outdated.

We renamed the `name` property to `value`:
https://github.com/dpfaffenbauer/CoreShop/commit/1f96fac24cf6ce19d37112433d3adde4e6529d44#diff-656b7a478de852d4b715106ac49f7e0ade545a002ef0930adf834970b7447b1fR20

And the `openButtonWidget`-method requires a config even tho it's not needed in that case:
https://github.com/dpfaffenbauer/CoreShop/commit/1f96fac24cf6ce19d37112433d3adde4e6529d44#diff-acdb71511c8fe828ede9157419f8118fd8ee883430f122b8db6b2f1bc88ea38dR9

``` typescript
       widgetManager.openBottomWidget({
            name: 'My first widget',
            component: 'my-first-widget',
            config: {}
        });
```

I will fix those cases in the next days in this PR with your changes included.
And will close it afterwards.

And maybe some additional information that could be interesting for you:

1. The type interpretation of the npm package was not working, since we had a small issue in our `package.json`.
This is fixed by now and should help you spot these errors.

2. Somewhere in the weeks till inspire we will revisit the plugin system - there is no fixed date, yet.
Than we will add a lot more of our UI components and and also more entrypoints for extension to our SDK.
So, just that you know. :)

dpfaffenbauer:
❤️ 

Nothing more I can say :)


Re 2. We would be happy to assist in all sorts possible, I guess we have quite some plugins to migrate and we, together, can learn a lot and enhance a lot.

markus-moser:
@dpfaffenbauer We needed to change the way how the entry points are loaded - therefore you would also need to adjust it in your core shop example:

https://github.com/pimcore/studio-ui-bundle/pull/1159

Issue #180, opened 2024-03-29
[User Management] User Editor


fashxp:
![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/1591ad44-6fd9-4590-851b-e4ec03fd8b20)

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/c8bbbccd-0a99-43a2-b877-2dc3b0558397)

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/033bcba5-04ab-4e0c-89f8-1a060aae5a4f)

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/041f5017-b43f-4ba2-9484-b3c49574316f)

PR #872, opened 2025-01-14
[User Management] Implement the User Roles entity

sholzer:
Task: https://github.com/pimcore/studio-ui-bundle/issues/180


assets/js/src/core/modules/user/roles/detail/tabs/settings/settings-container.tsx
```diff
@@ -0,0 +1,91 @@
...
+      form.setFieldsValue({
+        name: item?.name
+      })
+    }
+  }, [item, isLoading])
+
+  const onValuesChange = useCallback(
+    debounce((changedValues, allValues) => {
+      if (changedValues.permissionsDefault !== undefined || changedValues.permissionsBundles !== undefined) {
+        allValues.permissions = [
```
ValeriaMaltseva: In the current implementation, we modify the **allValues** object directly. While this works, it can sometimes lead to side effects, especially if **allValues** is tied to React's state or similar structures that expect immutability.

To ensure we avoid any unpredictable behavior, I propose creating a copy of **allValues** before making any changes.

assets/js/src/core/modules/user/roles/detail/tabs/workspaces/workspaces-container.tsx
```diff
@@ -0,0 +1,194 @@
...
+  ]
+
+  const objectsAccordion = [
+    {
+      key: '1',
+      title: <>{ t('user-management.workspaces.objects') }</>,
+      info: <> <IconTextButton
+        icon={ { value: 'add-find' } }
+        onClick={ () => {
+          setObjectWorkspaces([...item.dataObjectWorkspaces, {
```
ValeriaMaltseva: Since the code is quite similar in 3 places, what do you think about creating a function (for example, handleSetWorkspace or something similar) and passing the type as a prop? Inside the function, we could handle the logic dynamically based on the type.

assets/js/src/core/modules/user/roles/hooks/use-roles-draft.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  function removeItemFromState (): void {
+    if (item === undefined) return
+
+    dispatch(removeItem(item.id))
+  }
+
+  function changeItemInState (changes: any): void {
+    console.log('changeItemInState', changes)
```
ValeriaMaltseva: Does it need to be deleted? 😊

assets/js/src/core/modules/user/roles/hooks/use-roles-helper.tsx
```diff
@@ -0,0 +1,172 @@
...
+  RoleGetByIdApiResponse
+} from '@Pimcore/modules/user/roles/roles-api-slice.gen'
+
+interface IAddItemArgs {
+  parentId: number
+  name: string
+}
+
+interface IUseRoleReturn {
+  openItem: (id: number) => void
```
ValeriaMaltseva: I noticed that for the Role entity, we are using the name **_item_** instead of **_role_**. Could you please clarify the reasoning behind this naming choice?

In my opinion, since we already have many entities, it might be beneficial to use more specific and consistent naming conventions, such as role, to avoid confusion in cases where multiple entities are involved in the functionality.

What are your thoughts on this approach?

ValeriaMaltseva (changes requested):
Hi! The PR looks great already. I've just added a few comments to make it perfect. Please check them 😊

assets/js/src/core/modules/user/roles/detail/tabs/settings/settings-container.tsx
sholzer: done :)

assets/js/src/core/modules/user/roles/detail/tabs/workspaces/workspaces-container.tsx
sholzer: done :)

assets/js/src/core/modules/user/roles/hooks/use-roles-draft.tsx
sholzer: done :)

assets/js/src/core/modules/user/roles/hooks/use-roles-helper.tsx
sholzer: no idea why I did it that way :laughing: 
i renamed item to role :smiley:

ValeriaMaltseva (approved):
Thanks for fixing everything 😊

PR #878, opened 2025-01-16
Icon adjustments

xIrusux:
Resolves #

## Additional info
Focus areas:

image

image

image

image

image

image

image

image

image


xIrusux:
Added new filled icons and colored them. 
Will discus with ida if the exact shade needs color adjustment. 
image

xIrusux:
Update after the latest review with idea:

Something looks off here, which Ida is looking into:
image

Arrows need adjustments:
image

Issue #823, opened 2024-12-11
[Asset Folder Preview] Add batch options from list view


markus-moser:
In the asset folder list view we have this selection widget in combination with the batch operations (batch edit, export, zip download):

![Image](https://github.com/user-attachments/assets/deda976d-f4de-4e6c-bb92-4005ef71cf1c)

Currently it is missing in the folder preview. Please add it there exactly the same way and check that we do not need to duplicate code to achieve this:

![Image](https://github.com/user-attachments/assets/2a45191c-2c58-44fa-b95b-2455c37cbc32)

markus-moser:
This is not easily possible and also not available in the classic bundle, there for we skip it (at least for now) and remove the checkboxes.

PR #920, opened 2025-01-23
remove checkbox from preview view

xIrusux:
Resolves #


xIrusux:
After addressing [#823](https://github.com/pimcore/studio-ui-bundle/issues/823)
and adding the "All selected" and "Apply to selections" menu to the preview view we found that this only complicates the UI for the user as for example "batch edit" makes most sense in the context of the grid columns.
There also already exists a "export & zip download" for all assets in a folder on the preview view. 

We decided to remove the Checkboxes in the preview view for now to avoid further confusion.

Issue #293, opened 2024-05-21
Data Object Widgets - Versions


fashxp:
- [ ] allow detaching
- [ ] Single Version

for ruleset for flattening the data object layouts, see https://github.com/pimcore/product-management/issues/491#issuecomment-2110149208

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/6cd74b3d-38ed-43f8-9bb5-47b31a3e83fd)

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=3057-151818&t=090OJTL3MIaVURas-4

- [ ] Version Compare

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/12bc7934-f8bd-4b59-a944-a0be60f8fc66)

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=3274-159086&t=090OJTL3MIaVURas-4

ValeriaMaltseva:
In this task, the main logic of the Version Tab was implemented, including data preparation, rendering, and keeping it in one component that handles two types of elements: asset and data-object.

In a [separate task](https://github.com/pimcore/studio-ui-bundle/issues/987), all dynamic fields will be updated, and the view will be finalized.

PR #953, opened 2025-02-05
Data Object Widgets - Versions

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/293


assets/js/src/core/components/versions-fields-list/components/data-component/data-component.tsx
```diff
@@ -0,0 +1,56 @@
...
+        type="warning"
+      />
+    )
+  }
+
+  const objectDataType = objectDataRegistry.getDynamicType(currentFieldType)
+
+  const _props = {
+    ...props,
+    noteditable: true
```
markus-moser: Wouldn't it be a better idea to set this noteditable: true directly in the default getVersionObjectDataComponent() of the abstract class instead?

assets/js/src/core/modules/element/dynamic-types/defintinitions/objects/data-related/types/abstract/dynamic-type-object-data-abstract-input.tsx
```diff
@@ -33,4 +33,18 @@ export abstract class DynamicTypeObjectDataAbstractInput extends DynamicTypeObje
       />
     )
   }
+
+  getVersionObjectDataComponent (props: InputProps): React.ReactElement<AbstractObjectDataDefinition> {
```
markus-moser: Why do we need to override this method here? The abstract-input extends the regular input. Shouldn't be then our default implmentation from the `DynamicTypeObjectDataAbstract` be enough?

assets/js/src/core/modules/element/dynamic-types/defintinitions/objects/data-related/types/abstract/dynamic-type-object-data-abstract-numeric.tsx
```diff
@@ -53,17 +64,27 @@ export abstract class DynamicTypeObjectDataAbstractNumeric extends DynamicTypeOb
...
       min: fixUnsigned(props.unsigned === true, props.minValue, false) ?? undefined,
+      defaultValue: props.value,
       precision: props.integer === true ? 0 : (props.decimalPrecision ?? undefined),
-      step: props.increment ?? undefined
+      step: props.increment ?? undefined,
+      className: props.className
     }
   }
 
+  getVersionObjectDataComponent (props: AbstractNumericObjectDataDefinition): React.ReactElement<AbstractObjectDataDefinition> {
```
markus-moser, line 79: Why do we need to override this method here? The abstract-input extends the regular input. Shouldn't be then our default implmentation from the `DynamicTypeObjectDataAbstract` be enough? In my opinion we do not need any changes in this file?

markus-moser (changes requested):

assets/js/src/core/components/versions-fields-list/components/asset-versions-fields-view/asset-versions-fields-view.tsx
```diff
@@ -18,15 +18,15 @@ import { Flex } from '@Pimcore/components/flex/flex'
 import { Text } from '@Pimcore/components/text/text'
 import {
   type CategoriesList,
-  type IVersionsFieldsList,
+  type IAssetVersionsFieldsList,
```
markus-moser, line 21: Shouldn't this assets field view be in modules instead of components?`

assets/js/src/core/components/versions-fields-list/components/data-component/data-component.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
markus-moser: This should be in modules...

assets/js/src/core/components/versions-fields-list/components/object-versions-fields-view/object-versions-fields-view.tsx
```diff
@@ -0,0 +1,117 @@
+/**
```
markus-moser: This should be in modules...

assets/js/src/core/components/versions-fields-list/hooks/use-version-data.ts
```diff
@@ -0,0 +1,58 @@
+/**
```
markus-moser: This should be in modules...

assets/js/src/core/modules/data-object/editor/toolbar/toolbar.tsx
```diff
@@ -12,96 +12,75 @@
...
   const componentRegistry = container.get<ComponentRegistry>(serviceIds['App/ComponentRegistry/ComponentRegistry'])
   const ContextMenu = componentRegistry.get('editorToolbarContextMenuDataObject')
 
   useEffect(() => {
-    if (isSuccess && isSchedulesSuccess) {
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      messageApi.success(t('save-success'))
-      removeTrackedChanges()
-      resetModifiedDataObjectAttributes()
+    const handleSuccessEvent = async (): Promise<void> => {
```
markus-moser: Could you please revert all not version related changes in this file? I currently work on refactoring the save logic in this PR and we will then get some merge conflicts:

https://github.com/pimcore/studio-ui-bundle/pull/983

So we need to add the improvements on the new location then.

assets/js/src/core/modules/data-object/editor/toolbar/toolbar.tsx
```diff
@@ -116,28 +95,63 @@ export const Toolbar = (): React.JSX.Element => {
...
       }
     })
 
     const saveSchedulesPromise = saveSchedules()
 
-    Promise.all([saveDataObjectPromise, saveSchedulesPromise]).catch((error) => {
+    Promise.all([saveDataObjectPromise, saveSchedulesPromise]).then(() => {
+      const invalidateVersionsList = invalidatingTags.ASSET_VERSIONS(id)
+
+      dispatch(api.util.invalidateTags(invalidateVersionsList))
```
markus-moser: We should not do this here but implement it directly in the RTK query tags logic. Let's check it together with @xIrusux as she implemented the same approaches also for assets/other tabs.

assets/js/src/core/modules/element/dynamic-types/defintinitions/objects/data-related/components/object-brick/object-brick.tsx
```diff
@@ -33,12 +34,12 @@ export const ObjectBrick = (props: ObjectBrickProps): React.JSX.Element => {
   const currentValue = Form.useWatch(props.name, form)
 
   const dropdownItems: DropdownMenuProps['items'] = props.allowedTypes.map((type) => {
-    const hasType = currentValue?.some((fieldValue) => fieldValue.type === type) ?? false
+    const hasType = (isArray(currentValue) && currentValue.some((fieldValue) => fieldValue.type === type)) ?? false
```
markus-moser, line 37: Ok for now but @vin0401 needs to check the object bricks logic in general - data structure from API and frontend seem to not match anymore

assets/js/src/core/modules/element/dynamic-types/defintinitions/objects/data-related/types/abstract/dynamic-type-object-data-abstract-input.tsx
```diff
@@ -26,10 +26,12 @@ export abstract class DynamicTypeObjectDataAbstractInput extends DynamicTypeObje
     return (
       <Input
         autoComplete="off"
+        className={ props.className }
```
markus-moser: Not sure where the className prop here is initially defined? In general those class names should be defined more in the wrapper of the concret component (as discussed based on the `getObjectDataFormItemProps` method of the dynamic type class.

assets/js/src/core/modules/element/dynamic-types/defintinitions/objects/data-related/types/abstract/dynamic-type-object-data-abstract-numeric.tsx
```diff
@@ -45,25 +45,27 @@ const fixUnsigned = (unsigned: boolean, value: number | null, max: boolean): num
...
-    )
-  }
-
   getObjectDataComponentProps (props: AbstractNumericObjectDataDefinition): IInputNumberProps {
     return {
       inherited: props.inherited,
       disabled: props.noteditable === true,
       max: fixUnsigned(props.unsigned === true, props.maxValue, true) ?? undefined,
       min: fixUnsigned(props.unsigned === true, props.minValue, false) ?? undefined,
+      defaultValue: props.value,
```
markus-moser: We need to solve this differently - we cannot use defaultValue as this will break the data object editor handling. If there is no other way then changing the defaultValue we would need to do this in the `getVersionObjectDataComponent`

markus-moser (changes requested):

markus-moser:
@ValeriaMaltseva In general it looks already pretty good and we could merge it soon and do the missing parts in separate PRs, but please check my comments 👍

assets/js/src/core/components/versions-fields-list/hooks/use-version-data.ts
ValeriaMaltseva: It's a common hook, and I've updated the naming to show it. Thanks.

markus-moser (approved):

PR #982, opened 2025-02-07
Data objects refresh tabs

### LINKED ISSUE #955: [Data Object Versions] Check RTK Query Tags
Currently the versions tab only refreshes for assets but not for data objects. Please check the tags setup for data objects...

xIrusux:
Resolves #955

## Additional info
Here comes the first batch of RTK query updates. 
Here I am ensuring the version tab functions as expected. 

i.e. 

- [x] refreshes automatically when tab is closed and reopened through { refetchOnMountOrArgChange: true } option
- [x] refresh upon clicking the refresh Icon bottom left (this happens when navigating and therefore reopening the versions tab -> see last point)
- [x] reloads only the cached versions against the element that we are deleting / updating / publishing versions too 

Therefore the manual invalidation of tags is no longer needed. 

Let me know if you have any questions


xIrusux:
Already discussed with Valeriia - there is one test outstanding.

ValeriaMaltseva (approved):

PR #1000, opened 2025-02-12
Rtk tag improvements

### LINKED ISSUE #978: Check RTK query element tabs tags setup for data objects
Do all tabs refresh correctly?

xIrusux:
Resolves #978 

## Additional info

Part of this PR:

- [x] ensure element specific PUT, POST only invalidates the element specific queries that are related to the update. Using the ElementSpecificTag.

- [x] ensure DELETE only invalidates the tab specific "...DETAIL" Tag (i.e. SCHEDULE_DETAIL) which is then tied to both a schedule Id and the element specific query.
DELETE's API calls are id specific (i.e. scheduleId) but not element specific.
Without the tie to the element we would invalidate all schedules. This meant more than one open element tab would refresh their cached data. 

- [x] adding the ElementDetailTag to every Element call ensures this cached data is invalidated on an element refresh 

- [x] replaced instances of manual tag invaldiation with the correct invalidation logic as part of the query itself 

- [x] replaced instances of manual tag invalidation as part of useEffects on first load with the addition: ", { refetchOnMountOrArgChange: true })" to the query which has the same effect but avoids the occasional double loading that happened as a side effect previously

- [x] implemented coherent RTK tagging across the different Tabs for all currently existing API calls (also those that are not yet in use) 

- [x] removed unnecessary generic providing tags - because invalidating a generic tag automatically also includes id specific tags. which makes them obsolete. 
(this change below)
  PROPERTY_DETAIL: (id: string) => [{ type: tagNames.PROPERTIES, id }, tagNames.PROPERTIES],
  =>
    PROPERTY_DETAIL: (id: string) => [{ type: tagNames.PROPERTIES, id }],


xIrusux:
I have removed the { refetchOnMountOrArgChange: true } as it is no longer needed since we clear all tags on open. 
It is only still in place in the tags-tree. 
This is as I assumed we want to keep refreshing the configured tags on tag tab open. AVAILABLE_TAGS are not tied to an element so need to be triggered manually. 

One related thing I noticed was that we reload the asset tree when we save and publish properties for example. 
I am wondering if this is necessary?


____


FYI 
using the redux logger is awesome when debugging RTX Tags  
one just has to add it to the store core/app/store/index.ts 
and it logs every query invalidation / refetch based on tags. 

Screenshot 2025-02-13 at 14 22 27

assets/js/src/core/modules/element/editor/shared-tab-manager/tabs/tags/components/tags-tree/tags-tree-container.tsx
```diff
@@ -39,7 +39,7 @@ export const TagsTreeContainer = (props: TagsTreeContainerProps): React.JSX.Elem
     page: 1,
     pageSize: 9999,
     filter
-  })
+  }, { refetchOnMountOrArgChange: true })
```
markus-moser: Do we still need this?

markus-moser (approved):
I have the feeling that you did a great job here :)

I did some quick tests and did not find any problems. But as it is very complex to test it I will talk with @heigpa so that he maybe also can assist with testing. But please merge it (after checking my 1 single comment).

markus-moser:
> One related thing I noticed was that we reload the asset tree when we save and publish properties for example.
> I am wondering if this is necessary?

This is not really needed but we cannot know when saving what changed. Let's keep it like that for now and see how it feels.

assets/js/src/core/modules/element/editor/shared-tab-manager/tabs/tags/components/tags-tree/tags-tree-container.tsx
xIrusux: Hi Markus, 

I tried to explain this in my last comment.

This is as I assumed we want to keep refreshing the configured tags regularly. AVAILABLE_TAGS are not tied to an element so they do not automatically get refreshed as part of our element refresh functionality. 

They only get invalidated when one adds, removes or creates a new tag in the Tag Configuration. 

However, other users may also add or remove tags in the configuration so I think we should ensure that query refreshes on some regular basis. 

What do you think?

xIrusux:
that sounds good re the asset tree, lets keep an eye on it

xIrusux:
To make things a bit cleaner I have instead moved the AVAILABLE_TAGS invalidation into the tag invalidation of the **element detail tag** that way going forward we can try to not control refetching manually without an exception.
Screenshot 2025-02-14 at 16 47 02

assets/js/src/core/modules/element/editor/shared-tab-manager/tabs/tags/components/tags-tree/tags-tree-container.tsx
markus-moser: Aaah, I see - yes, in that case we should keep it

Issue #981, opened 2025-02-07
[Error Boundary] Deduplicate or debounce error modals


markus-moser:
Sometimes it's the case that a component gets re-rendered multiple times within a very short period of time. E.g. the `<StrictMode>` which is enabled in our dev build renders each component twice by default. But there can also be other constellations where components are re-rendered several times. In that case if an error (e.g. ApiError) is triggered the error modal is shown multiple times with the same message (so when clicking ok the message appears again with the same message).

Can we add a mechanism to avoid this?

Another option would be some kind of debouncing mechanism so that we only show maximum 1 error within 1 second or so and skip the others.

PR #1001, opened 2025-02-12
[Error Boundary] Fix the bug of duplicate  error messages

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/981


markus-moser:
@ValeriaMaltseva I tried it out. It solves the problem to some extend. But I have a constellation where the error still pops-up multiple times as it not triggered within one rendering cycle but the component is re-rendered twice (and then again twice because of the strict mode). Not sure how we can tackle that.

markus-moser:
@ValeriaMaltseva Aaah, I checked my case again and found out that it is really a very specific edge case which is not relevant for the global error handler. Fine, so this topic is solved 👍

markus-moser (approved):

Issue #979, opened 2025-02-07
Image preview does not get refreshed after editing an image


heigpa:
**Steps to reproduce**

- Login as admin
- Open an image
- Edit the image
- Save the edited image
- Open the preview

**Actual behaviour**

- Changes should be seen on the preview page
- Refresh button has no effect
- Changes are seen after a browser refresh

**Expected behaviour**

- Changes should be seen immediately in the preview tab

**Environment**

- Studio nightly
- Windows
- Google Chrome

heigpa:
Does not work on nightly

ValeriaMaltseva:
@heigpa 
I rechecked, and it works.

P.S. Please make sure to save it by clicking the "Save" button

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/7100784f-0e78-400f-b958-b38c0d70f161" />

heigpa:
@ValeriaMaltseva it works now. Thank you!

PR #1007, opened 2025-02-14
[Assets] Fix the issue of refreshing the image after editing and saving it

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/979


templates/image-editor/editor.html.twig
```diff
@@ -160,12 +160,21 @@
...
+
             fetch(url, {
                 method: 'PUT',
                 headers: {
                     'Content-Type': 'application/json',
                 },
                 body: JSON.stringify({ data: { dataUri } })
+            }).then((response) => {
+                if (response.ok) {
+                    // Post a message to the Preview Asset Tab indicating that the image was saved
```
xIrusux: Cool way of doing this 👍 . 
I wonder if using a object would be more helpful here in case you want to extend this in the future and to make it easier to parse / easier to debug as that would allow filtering based on type for example.

window.parent.postMessage(
    { type: "MiniPaint", status: "success", message: "Image successfully saved!" },
    isDev ? devTargetOrigin : prodTargetOrigin
);

assets/js/src/core/modules/asset/editor/types/image/tab-manager/tabs/preview/preview-view.tsx
```diff
@@ -23,17 +23,38 @@ interface PreviewViewProps {
...
-  const { styles } = useStyle()
   const { src } = props
+
+  const [imageSrc, setImageSrc] = useState(src)
+
+  const { styles } = useStyle()
   const { zoom, setZoom } = React.useContext(ZoomContext)
 
+  useEffect(() => {
+    // Event handler to handle the message event
```
xIrusux: Hm - I am wondering how we want to handle commenting our code in general. 
Here I find that the comment is stating something quite obvious. What do you think?

assets/js/src/core/modules/asset/editor/types/image/tab-manager/tabs/preview/preview-view.tsx
```diff
@@ -23,17 +23,38 @@ interface PreviewViewProps {
...
+    const handleMessage = (event: any): void => {
+      if (event.data === IMAGE_SAVE_SUCCESS_MESSAGE) {
+        // Update the image source by adding a query parameter to force reloading of the image
```
xIrusux: Here the "forcing to reload" seems to be the important part. I would usually try to incorporate that in a function name to avoid comments.

xIrusux (commented):
Looks good to me :). Just a few small notes.

xIrusux (approved):

PR #1010, opened 2025-02-18
Global notes events

### LINKED ISSUE #161: Global Notes & Events view
Paginated Notes Listing with filter and fieldFilters: http://localhost/pimcore-studio/api/docs#/Notes/note_get_collection

(Maybe we should switch this to a post request, since we send json in the query for the fieldFilters)

Delete notes
http://localhost/pimcore-studio/api/docs#/Notes/note_delete_by_id

### LINKED ISSUE #979: Image preview does not get refreshed after editing an image

xIrusux:
Resolves #161

## Additional info

First PR for global notes and events container 

https://github.com/user-attachments/assets/5064b2c2-bbb5-4b35-bafb-16bf2a8df3c7

Next: 
- shared component for this layout 
- allow showing different kinds of details data


assets/js/src/core/modules/notes-and-events/notes-and-events-container.tsx
```diff
@@ -0,0 +1,93 @@
...
+              gap={ 'medium' }
+              vertical
+            >
+              <Flex
+                justify={ 'space-between' }
+              >
+                <Flex
+                  gap={ 'small' }
+                >
+                  <Title>Notes & Events</Title>
```
ValeriaMaltseva: Should we translate this?

assets/js/src/core/modules/notes-and-events/table/note-modal.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <>
+        {displayTextArea(noteDetail.type, t('notes-and-events.columns.type'))}
+        {displayTextArea(noteDetail.title, t('notes-and-events.columns.title'))}
+        {displayTextArea(noteDetail.description, t('notes-and-events.columns.description'), false)}
+        {displayTextArea(noteDetail.fields.toString(), t('notes-and-events.columns.details'))}
+        {isString(noteDetail.userName) && displayTextArea(noteDetail.userName, t('notes-and-events.columns.user'))}
+        <Box
+          margin={ 'small' }
+        >
+          <div>Date</div>
```
ValeriaMaltseva: Should we translate this?

assets/js/src/core/modules/notes-and-events/table/table.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const [notes, setNotes] = useState<DataNote[]>([])
+  const [noteDetail, setNoteDetail] = useState<DataNote | undefined>(undefined)
+
+  useEffect(() => {
+    if (notesAndEvents !== undefined && Array.isArray(notesAndEvents)) {
+      setNotes(enrichNotesAndEvents(notesAndEvents))
+    }
+  }, [notesAndEvents])
+
+  const enrichNotesAndEvents = (data: Note[]): DataNote[] => {
```
ValeriaMaltseva: Can just simplify 😉

```
const enrichNotesAndEvents = (data: Note[]): DataNote[] => (
    data.map((item) => ({
      ...item,
      fields: item.data.length,
      rowId: uuid()
    }))
  )
```

assets/js/src/core/modules/notes-and-events/table/table.tsx
```diff
@@ -0,0 +1,186 @@
...
+          <Flex
+            align='center'
+            className='w-full'
+            justify='center'
+          >
+            <IconButton
+              icon={ { value: 'open-folder' } }
+              onClick={ async () => {
+                !isUndefined(elementType) &&
+                                await openElement({
```
ValeriaMaltseva, line 145: What do you think about moving this into a separate function to avoid duplication?

ValeriaMaltseva (approved):
It looks pretty good already! 😊

I've added some points, but it is not a blocker, and you can fix them in the next PR regarding the Notes and Events.

assets/js/src/core/modules/notes-and-events/notes-and-events-container.tsx
xIrusux: uh that was an oversight. Thanks

assets/js/src/core/modules/notes-and-events/table/note-modal.tsx
xIrusux: same :)

assets/js/src/core/modules/notes-and-events/table/table.tsx
xIrusux: simplified

assets/js/src/core/modules/notes-and-events/table/table.tsx
xIrusux: hm I see what you mean. I saw it more as a is not null check and then the function call which is the same in 2 places. If there are more calls I will change it.

PR #1021, opened 2025-02-19
Notes and events details

xIrusux:
Part of #161

- refactored existing sub details table from notes-view 
- fixed loading the shape of "timeWorked" 
- implemented in new global notes and events modal

## Additional info

https://github.com/user-attachments/assets/9b201c84-03a6-466f-b9b2-8bf722a2c4f5


xIrusux:
Screenshot 2025-02-21 at 12 12 43

- Added some path handling / displaying it as a Tag with a resizable breadcrub
- clicking on the path opens it in a new window 
- other types of objects / strings are being handled with a fallback to a JSON.stringify 

As a next step I would like to write a small adapter to handle the breadcrumb component based on the different use cases (specifically the intended width)

ValeriaMaltseva (approved):

PR #1026, opened 2025-02-20
Fix the RGBA color data type in the versions view

### LINKED ISSUE #1008: [Versions] The value in the ColorPicker is always empty
**Steps to reproduce:**
1. Open the object that contains the Color field.
2. Change the color to any value.

Image

3. Pay attention to the data that the frontend sends to the backend.

Image
Image

4. Go to the Version tab.
5. Choose the Auto saved item.
6. The Color field is empty. Also, pay attention to the second screen: the value from the **_api/versions/{id}_** is not the same as in point 3 of this issue (with "ff" at the end).

Image
Image

**Actual behaviour**: the value in the ColorPicker is empty.
**Expected behaviour**: the value in the ColorPicker should be the same as the one we send in point 3.

markus-moser:
Resolves #1008


markus-moser:
@ValeriaMaltseva This PR should fix your reported issue. The color picker is able to handle the various different formats but we had the same problem like for other data types that we need to forward the value prop (as they are not wrapped within a <FormItem> in the versions view like in the data object editor)

ValeriaMaltseva (approved):

ValeriaMaltseva (approved):
@markus-moser I've checked the fix locally, and it works. Thanks a lot! 😊

PR #1029, opened 2025-02-20
Fixed refresh logic for assets

### LINKED ISSUE #963: Refresh problem when deleting an asset in list view
**Steps to reproduce**
- Login as admin
- Open the asset tree
- Open a folder in the list view
- Delete an asset with a right mouse click from the list

**Actual behaviour**
- The folder window changes its view to the 'Preview'
- The deleted asset is still being shown
- Refreshing the window does not work
- It needs a browser refresh

Image

**Expected behaviour**
- Asset is being deleted from the list view and the focus stays on the list view
- When changing to the preview view, the asset should also be deleted

**Environment**

- Studio nightly
- Windows
- Google Chrome

Corepex:
Resolves #963


Corepex:
I discovered the bug that if i delete an asset in the list view it always dispatches the delete folder action

image

markus-moser:
@Corepex Yes I also saw this - please create an issue for the backend team. Single items need to be deleted directly.

Corepex:
obsoleted by: https://github.com/pimcore/studio-ui-bundle/pull/1326

PR #1064, opened 2025-02-24
Content layout addition

xIrusux:
Resolves #

## Additional info

Added a search bar / space between version of the toolbar to the toolbar shared component in storybook and updated places of use accordingly 

image

Screenshot 2025-02-24 at 13 00 59
Screenshot 2025-02-24 at 13 01 12


assets/js/src/core/components/content/content.tsx
```diff
@@ -27,12 +27,15 @@ export interface ContentProps extends Omit<BoxProps, 'children'> {
   centered?: boolean
   fullPage?: boolean
   noneOptions?: INoContentProps
+  padding?: SizeDefinition
+  margin?: SizeDefinition
```
vin0401: Both should already be inferred by the extended type.
So they should not be needed to be repeated here?

assets/js/src/core/components/content/content.tsx
```diff
@@ -27,12 +27,15 @@ export interface ContentProps extends Omit<BoxProps, 'children'> {
...
 }
 
 export const Content = ({
   children,
   padded = false,
   padding = { top: 'small', x: 'extra-small', bottom: 'extra-small' },
+  margin,
```
vin0401: Let's set the default directly here, instead of line 65?

assets/js/src/core/components/toolbar/toolbar.tsx
```diff
@@ -23,20 +24,25 @@ export interface ToolbarProps {
...
     `toolbar--size-${size}`
-  ].join(' ')
+  ]
+
+  const paddingClasses: string[] = getSizingClasses('p', padding)
+  const marginClasses: string[] = getSizingClasses('m', margin)
 
   return (
-    <div className={ classes }>
+    <div className={ `${classes.join(' ')} ${paddingClasses.join(' ')} ${marginClasses.join(' ')}` }>
```
vin0401: Wouldn't the box component also work in this case? 
Since it's designed to manage these paddings and margins?

assets/js/src/core/components/toolbar/toolbar.tsx
```diff
@@ -23,20 +24,25 @@ export interface ToolbarProps {
   theme?: 'primary' | 'secondary'
   position?: 'top' | 'bottom'
   size?: 'small' | 'default'
+  padding?: SizeDefinition
+  margin?: SizeDefinition
```
vin0401: Would prefer a more explicit typing, when we use the one from the Box.
In case we are switching something just on the padding but not on the margin it would directly affect that component right here.

```
padding?: BoxProps['padding']
margin?: BoxProps['margin']
```

This should be safer.

vin0401 (changes requested):

vin0401 (approved):

PR #1087, opened 2025-02-27
Quick action open

### LINKED ISSUE #1004: Open asset/data object/document quick actions
Image

- Based on the related backend implementation (https://github.com/pimcore/studio-backend-bundle/issues/756)
- Add entries to the main menu (please discuss with @idaiv where they should be located exactly)
- Use `formModal` to ask the user to enter id/path
- Fetch backend API to request the actual object/asset(/document) ID
- Use existing helpers to open the element

xIrusux:
Resolves #1004


xIrusux:
image

assets/js/src/core/modules/app/nav/index.ts
```diff
@@ -21,6 +21,12 @@ moduleSystem.registerModule({
   onInit: () => {
     const mainNavRegistryService = container.get<MainNavRegistry>(serviceIds.mainNavRegistry)
 
+    mainNavRegistryService.registerMainNavItem({
```
markus-moser, line 24: I just moved the file menu to the top of the list so that it is the same for now like in the classic bundle ;)

assets/js/src/core/modules/open-element/hooks/open-element-helper.ts
```diff
@@ -0,0 +1,54 @@
...
+  openElementByPathOrId: (value: string | number | undefined, elementType: ElementType) => Promise<void>
+  isLoading: boolean
+}
+
+export const openElementHelper = (): OpenAssetHelperReturn => {
+  const { openElement } = useElementHelper()
+
+  const [fetchElementId, { isLoading }] = useLazyElementGetIdByPathQuery()
+  const handleFetchAndOpen = async (path: string, elementType: ElementType): Promise<void> => {
+    try {
```
markus-moser, line 33: We need to somehow improve the error handling here. At the moment when the element was not found I get 2 error modals with slightly different messages:

1. 
image

2.
image

Instead we should get a translated error message like in the classic bundle:

image

Please ask the backend team to provide a specialized errorKey for the case where the element was not found so that we can handle this case with a translated message.

assets/js/src/core/modules/open-element/hooks/open-element-helper.ts
```diff
@@ -0,0 +1,54 @@
...
+      const result: ElementGetIdByPathApiResponse = await fetchElementId({ elementType, elementPath: path }).unwrap()
+      await openElement({ id: result.id, type: elementType })
+    } catch (error) {
+      trackError(new GeneralError(`Error fetching element: ${get(error, 'data.message')}`))
+    }
+  }
+
+  const openElementByPathOrId = async (value: string | number | undefined, elementType: ElementType): Promise<void> => {
+    try {
+      if (!isNaN(Number(value))) {
```
markus-moser: We need to always ask the server here - also for numbers. The logic can be extended on backend side with events. So depending on the system setup it' not always the case that a number is a Pimcore ID.

assets/js/src/core/modules/open-element/hooks/open-element-helper.ts
```diff
@@ -0,0 +1,54 @@
...
+import { get } from 'lodash'
+interface OpenAssetHelperReturn {
+  openElementByPathOrId: (value: string | number | undefined, elementType: ElementType) => Promise<void>
+  isLoading: boolean
+}
+
+export const openElementHelper = (): OpenAssetHelperReturn => {
+  const { openElement } = useElementHelper()
+
+  const [fetchElementId, { isLoading }] = useLazyElementGetIdByPathQuery()
```
markus-moser: This is not the right endpoint. There is a specialized endpoint for this task provided by the backend which handles all cases:

`/elements/{elementType}/resolve'`

(https://github.com/pimcore/studio-backend-bundle/pull/772)

markus-moser (changes requested):
There are still some details which we need to solve - please check my comments ;)

assets/js/src/core/modules/open-element/hooks/open-element-helper.ts
xIrusux: Screenshot 2025-03-03 at 16 26 54

I have raised an issue to use a new error key 'error_element_not_found' for this api call.

https://github.com/pimcore/studio-backend-bundle/issues/867

https://github.com/pimcore/studio-ui-bundle/blob/fc53872f309a79131d607a15e03ff89352ff9a95/assets/js/src/core/modules/app/error-handler/classes/api-error.ts#L45

At the moment a GenericError errorKey is returned. 
Whenever that is the case we avoid finding the correct translation and instead look for a the message which is then more detailed. 

Perhaps also a solution that works here? Let me know what you think

markus-moser: The problem here is that the message is not translated. Therefore we should wait for the errorKey and then add it to our translations.

xIrusux: I will merge this version and update it once we have the specific keys. Thanks for clarifying

PR #1094, opened 2025-02-28
Global Notes and Events improvements

xIrusux:
Resolves #

## Additional info

- updated search to only fire on search button click 
- added search bar version with button to storybook
- updated pagination component to use type safety (stop doing number.toString) to avoid unexpected behaviour onPageSizeChange
- hide element tag + actions when there is no element 

Screenshot 2025-02-28 at 19 41 08


assets/js/src/core/modules/notes-and-events/table/table.tsx
```diff
@@ -132,32 +134,34 @@ export const Table = ({ notesAndEvents, notesAndEventsFetching }: TableProps): R
...
+                onClick={ async () => {
+                  !isUndefined(elementType) &&
+                                    await openElement({
+                                      type: elementType,
+                                      id: elementId
+                                    })
+                } }
+                type="link"
+              />
+              <IconButton
```
markus-moser, line 154: We should only hide the "open element" button but not the details modal when no element is found. In the details modal there could still be some relevant information.

markus-moser (changes requested):
Please check my single comment - other then that it LGTM 👍 

But please rename the PR - it's about "notes&events improvements" not about tag config.

assets/js/src/core/modules/notes-and-events/table/table.tsx
xIrusux: Of course. Thanks for pointing it out.

PR #1095, opened 2025-03-03
update searchbar to only search on button click & add tag text on rename

xIrusux:
Resolves #

## Additional info
Small change that ensures the search is update as per the global notes section (i.e. search only on button click)
+  the tag text is shown on rename


Corepex (approved):
LGTM

PR #1101, opened 2025-03-04
1096 search update

### LINKED ISSUE #1096: Search update
Working on the Notes & Events we realised that we would like to load the search results only when the user has intentionally hit enter. To simplify this integration, we decided to go back to the default Ant Design search input. 

Default Ant search for most of the cases:
https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=8997-265411&t=s0jsfZPm64sYXmlJ-4
 
The library component: https://www.figma.com/design/lomefPHRhJnkljB2nXtuKV/Ant-Design-5.11?node-id=515-41351&t=n446fD480KKfmEgk-4
 
Hover over states can be seen here: https://ant.design/~demos/input-demo-search-input

Image

 The only exception where version Default (on the screenshots of Storybook and here https://studionightly.pimcore.app/storybook/?path=/docs/components-controls-searchinput--docs) would be for now in the User Management and other such super tiny spaces where we remove the button simply to assure enough space for the user input (example: https://www.figma.com/design/pg65AmcoU5iVtQeroHK4iF/Full-screens?node-id=11063-259616&t=F5786IgVuZIqKu7C-4)

*Please apply the default style for the button as designed (the borders are with #ColorBorder on default and only when hovered they change to the purple ones).

xIrusux:
Resolves #1096

## Additional info

Screenshot 2025-03-04 at 13 55 54

- minimal search is used in the filter & user management


assets/js/src/core/modules/app/theme/utils/themes/default-theme.ts
```diff
@@ -100,6 +100,56 @@ const defaultTheme = {
       colorBgContainer: '#ffffff',
       borderRadiusSM: '4px'
     },
+    Input: {
```
vin0401: Do we need those? Shouldn't these be the default values from ant already?
Because when I get the diff in the theme buddy I don't see them.

vin0401 (changes requested):

assets/js/src/core/components/search-input/search-input.styles.ts
```diff
@@ -16,17 +16,28 @@ import { createStyles } from 'antd-style'
...
-    max-width: 300px;
-   
-      .ant-input-prefix {
-        margin-inline-end: ${token.marginXS}px;
+ 
+    .ant-input-prefix {
+    margin-inline-end: ${token.marginXS}px;
+    }
+     
+     :where([class*="css-dev-only-do-not-override-"]).ant-input-search
```
xIrusux: This is super ugly -,-
I am very open to other suggestions :)

assets/js/src/core/modules/tags/tag-configuration-container.tsx
```diff
@@ -201,7 +202,7 @@ const TagConfigurationContainer = (): React.JSX.Element => {
         }
     >
       <Content
-        loading={ tagsFetching }
+        loading={ tagsLoading }
```
xIrusux, line 205: this is to ensure the loading states upon tag config search look ok (was lost in a merge I believe)

assets/js/src/core/components/search-input/search-input.styles.ts
vin0401: This will not work in production mode. Since the css-dev-only... will not be available anymore.

vin0401 (approved):

Issue #987, opened 2025-02-11
[Versions] Update the fields view for the data-object type


ValeriaMaltseva:
To-Do:
- [x] Check and update the dynamic types for "simple" fields where needed.
- [x] Add the logic for the "Localized Fields" type.
- [x] Add the logic for the "Objects Bricks" type.
- [x] Add the logic for the "Field Collections" type.
- [x] Add the logic for the "Block" type.
- [x] Fix styles.

PR #1115, opened 2025-03-11
[Versions] Add new complex types

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/987

What has been done as part of this PR:
- Added the logic for the "Objects Bricks" type.
- Added the logic for the "Field Collections" type.
- Added the logic for the "Block" type.
- Updated the common logic for rendering the version layout.
- Updated styles.


markus-moser:
For the table data types we still don't have a (yellow) modified attribute indicator:

image

markus-moser:
I get this error when I try to compare the versions of a news object (which contains field collections):

image

markus-moser:
When I try to compare blocks I get the following screen:

image

It seems that for some reason this uses a massive amount of memory - I used only a very simple block:

image

markus-moser (approved):
Was not able to test it but please lets merge it and do fixes in follow-up PRs

Issue #1120, opened 2025-03-13
Do not call irrelevant endpoints in login form


markus-moser:
Some of the endpoints do not make sense on the login form

![Image](https://github.com/user-attachments/assets/5c376949-5aaf-4402-a629-a6c0df134b82)

Also currently when not logged in we get sometimes an error like this:

![Image](https://github.com/user-attachments/assets/40d55570-0e36-4aa7-9e28-1448b02c1faf)

PR #1146, opened 2025-03-18
Do not call irrelevant endpoints in login form

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/1120


assets/js/src/core/modules/app/app-loader.tsx
```diff
@@ -119,17 +122,29 @@ export const AppLoader = (props: IAppLoaderProps): React.JSX.Element => {
       })
   }
 
+  const loadUserData = (): void => {
+    Promise.all([
+      initSettings(),
+      initActivePerspective()
```
markus-moser: We still have a problem here. `initSettings`and `initActivePerspective` need to be executed in sequential order as the perspective logic uses a setting from the settings endpoint. Otherwise depending on the timing we can get this error:

image

Could you please add a logic so that we first load the settings and then the perspectives? Please prepare in a way so that we can also load other endpoints in parallel to the perspectives but after loading the settings.

markus-moser (changes requested):

markus-moser (approved):
Fine, now all problems seem to be solved 👍

PR #1158, opened 2025-03-20
Added `missingTranslations` store

### LINKED ISSUE #111: Translations Phase 1
- [ ] Consider login page translations
- [ ] Creating new translation entries for missing translations

Corepex:
Resolves #111 
Blocked by: https://github.com/pimcore/studio-backend-bundle/issues/930

## Additional info
This pull request introduces several changes to the translation handling system in the `assets/js/src/core/app/i18n` module. The changes focus on adding functionality to handle missing translations and updating the store configuration to support dynamic middleware.

### Translation Handling Enhancements:

* [`assets/js/src/core/app/i18n/helper/translation-helper.ts`](diffhunk://#diff-1d947444badc5b6f58d602560706b75853be440dfe2dff74c912885c129d288aR1-R28): Added a new helper function `addNewTranslations` to update translations through the store.
* [`assets/js/src/core/app/i18n/index.ts`](diffhunk://#diff-0b4024656b2848c25a63406f4f6a01f515bd196f315646228892954153fecbc9R16-R17): Implemented handling of missing translation keys by dispatching the `addMissingTranslation` action when a key is missing. [[1]](diffhunk://#diff-0b4024656b2848c25a63406f4f6a01f515bd196f315646228892954153fecbc9R16-R17) [[2]](diffhunk://#diff-0b4024656b2848c25a63406f4f6a01f515bd196f315646228892954153fecbc9L32-R34)
* [`assets/js/src/core/app/i18n/store/missingTranslations.slice.ts`](diffhunk://#diff-35c7e2df11f62b515d6b48dbdbfa9d5fd2cd332c53d19c4511c9fb63769fbc95R1-R54): Created a new Redux slice `missingTranslations` to manage missing translation keys, and added middleware to debounce and send missing translations after a period of inactivity.

### Store Configuration Updates:

* [`assets/js/src/core/app/store/index.ts`](diffhunk://#diff-8199c328338575d91d2477e9e146a3017ad30fa9d043636c33b35314e9f550cbL14-R14): Added dynamic middleware capabilities to the store configuration, allowing middleware to be added dynamically. This includes the `createDynamicMiddleware` function and related methods. [[1]](diffhunk://#diff-8199c328338575d91d2477e9e146a3017ad30fa9d043636c33b35314e9f550cbL14-R14) [[2]](diffhunk://#diff-8199c328338575d91d2477e9e146a3017ad30fa9d043636c33b35314e9f550cbR37-R42) [[3]](diffhunk://#diff-8199c328338575d91d2477e9e146a3017ad30fa9d043636c33b35314e9f550cbL49-R55) [[4]](diffhunk://#diff-8199c328338575d91d2477e9e146a3017ad30fa9d043636c33b35314e9f550cbR73-R75)


assets/js/src/core/app/i18n/index.ts
```diff
@@ -29,7 +31,7 @@ i18n
   })
 
 i18n.on('missingKey', (lngs, namespace, key, res) => {
-// @todo implement handling of missing keys after endpoints are available
+  store.dispatch(addMissingTranslation(key))
```
vin0401, line 34: Shouldn't we check for the language here as well?
Right now, we only sending english translation key to the server.

But how does this behave when there is a missing key only in e.g. german?
Will it override the english value then? 

I'm also not sure how often this missingKey Event is called.
Can it report the same missing key twice?
For example when you reopen an widget?

assets/js/src/core/app/i18n/helper/translation-helper.ts
```diff
@@ -0,0 +1,28 @@
...
+*
+*  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import { store } from '@Pimcore/app/store'
+import { api as translationApi } from '@Pimcore/modules/app/translations/translations-api-slice.gen'
+
+export const addNewTranslations = async (translations: string[]): Promise<void> => {
+  await store.dispatch(translationApi.endpoints.translationUpdate.initiate({
```
vin0401, line 18: Just for my interest, how does the endpoint behave, when there is already an available translation on the server.
Will it still override the value?

Corepex: This is a good question - I will double-check that with the BE team

assets/js/src/core/app/i18n/index.ts
Corepex: This shouldn't be a problem since the `missingKey` only gets called if the key is absent. 
That means that it's not there in any language ... 

If we have, for example, a new `foo.bar` key in the UI, it gets added to the backend and, with that, also added to all languages as '' (default empty string). From that moment on, it's there—empty but there.

I just noticed that we should update the translations after pushing a new key. 
I will double-check that with the BE team ...

Corepex: I just double checked the docs. I18n adds missing translations to the fallback language. So it doesn't hit twice ...

vin0401: Okay, then I think we are fine in this regard.

assets/js/src/core/app/i18n/helper/translation-helper.ts
vin0401: As discussed, we should have something like a create endpoint for the missing translations.
Just to ensure, that we not override a translation, that already exists. 
The endpoint should also fail silently in case of an existing key. This key should then be skipped.

vin0401 (approved):

Corepex:
Created BE ticket for followup https://github.com/pimcore/studio-backend-bundle/issues/978

PR #1159, opened 2025-03-20
Change the way webpack entry points are loaded

markus-moser:
To make it easier to make studio optional in Pimcore bundles we needed to refactor the way webpack entry points/static resources are loaded within Pimcore bundles. Take a look at the updated docs or at this commit for an example:

https://github.com/pimcore/quill-bundle/pull/23/commits/2b9edc26f909deb0764bd598cba97e9ad6d08d8e


markus-moser:
Failed tests are not releated to this PR, therefore I ignore them for now

PR #1174, opened 2025-03-24
translate main nav items

### LINKED ISSUE #1084: [Main Nav] Add translation keys for existing menu entries
Currently the labels are hardcoded

xIrusux:
Resolves #1084 

## Additional info
With the way how modules are currently set up, using the Context of the translation poses an invalid hooks call when naming widgets i.e. windows or the base menu items (i.e. "Tools", "File", "Settings")

It may be worth looking into another way of doing this.


assets/js/src/core/modules/app/nav/main-nav.tsx
```diff
@@ -177,31 +177,31 @@ export const MainNav = (): React.JSX.Element => {
                   <ul className={ 'main-nav__list-inline' }>
                     <li><IconTextButton
                       icon={ { value: 'pimcore' } }
-                        >Default</IconTextButton></li>
+                        >{t('navigation.default')}</IconTextButton></li>
                     <li><IconTextButton
                       icon={ { value: 'cdp' } }
                       type={ 'default' }
-                        >CDP</IconTextButton></li>
+                        >{t('navigation.cdp')}</IconTextButton></li>
```
markus-moser: We don't need to translate the perspectives (as this is only a static example prepared in the template. It depends on the user configuration which perspectives we will have.

We need to translate the main nav entries itself (like "Tools", "Notes&Events" etc.). Yes, in the module Calling the `t()` method will result in an error, but we can just pass the label as translation key there and then call the `t()` function in the template:

https://github.com/pimcore/studio-ui-bundle/blob/5d394b3c0475576f1b39b5e57bb62339f32babcd/assets/js/src/core/modules/app/nav/main-nav.tsx#L86

markus-moser (changes requested):

assets/js/src/core/modules/widget-manager/title/tab-title-container.tsx
```diff
@@ -43,7 +45,7 @@ export const TabTitleContainer = ({ node, modified }: TabTitleContainerProps): R
     return (
       <BorderTitleView
         icon={ icon }
-        title={ node.getName() }
+        title={ t(`${node.getName()}`) }
```
markus-moser, line 50: With this approach we need to be a bit careful. I think we need an additional attribute to make it configurable in the widget config if we want to translate the tab title or not. For your cases this is totally fine, but this is also used for our assets/data object(/document) tabs and if we translate them we create a translation key for each single element. So if we have e.g. 1.000.000 of assets with different file names in our system we would potentially produce 1.000.000 of useless translation keys.

markus-moser (changes requested):
Please check my comment, other then that it LGTM 👍

assets/js/src/core/modules/widget-manager/title/tab-title-container.tsx
xIrusux: fixed :)

markus-moser: Not sure if I get it but I still see the same line of code?

xIrusux:
new clean PR - https://github.com/pimcore/studio-ui-bundle/pull/1283

PR #1207, opened 2025-03-27
feat(error-handler) add title option for error handler modal

### LINKED ISSUE #1188: [Data Object Editor] improve validation error messages
**Steps to reproduce**
- Login as admin
- Open a data object with types with a configured validation
- Maintain the data object and violate one of the validations

**Actual behaviour**
- General error message is shown

Image

**Expected behaviour**
- Error message should be meaningful to see which error appeared for which field

**Environment**

- Studio nightly
- Windows
- Google Chrome

sholzer:
Resolves #1188


markus-moser:
@ValeriaMaltseva Could you please review it? The goal was to implement a special case for those validation errors. In that case we should have the errorKey translation text as translated title of the modal and the message in the body.

assets/js/src/core/components/modal/alert-modal/hooks/use-alert-modal.tsx
```diff
@@ -36,27 +37,29 @@ export const useAlertModal = (): UseAlertModalResponse => {
 
   return useMemo<UseAlertModalResponse>(
     () => ({
-      info: ({ content }) => (
+      info: ({ title, content }) => (
         modal.info({
-          title: t('info'),
+          title: title !== undefined ? t(title) : t('error'),
```
ValeriaMaltseva: Should we use **t('info')** instead of t('error') here?

assets/js/src/core/components/modal/alert-modal/hooks/use-alert-modal.tsx
```diff
@@ -36,27 +37,29 @@ export const useAlertModal = (): UseAlertModalResponse => {
...
-        modal.error({
-          title: t('error'),
-          content
-        })
-      ),
-      warn: ({ content }) => (
+      error: ({ title, content }) => {
+        return (
+          modal.error({
+            title: title !== undefined ? t(title) : t('error'),
```
ValeriaMaltseva: What do you think about replacing the logic with: **!isUndefined(title)**: t(title) : t('error')?

assets/js/src/core/modules/app/error-handler/error-handler.tsx
```diff
@@ -52,7 +52,7 @@ const trackError = (data: IErrorContentProvider, handler?: ErrorHandler): never
     handler(getErrorContentValue())
   } else {
     // default handler
-    ErrorModalService.showError(getErrorContentValue())
+    ErrorModalService.showError(getErrorContentValue(), typeof errorContent === 'object' ? errorContent.title! : null)
```
ValeriaMaltseva: What do you think about updating this logic to:
ErrorModalService.showError({ content: getErrorContentValue(), title: errorContent?.title }) 

1. This way, we don't need to worry about the correct order of arguments
2. Also, _typeof errorContent === 'object'_ feels a bit overwhelming to me...Or am I missing a specific reason for it?

ValeriaMaltseva (changes requested):
Hey @sholzer!

The general approach looks good, but I've left a few comments. Please have a look.

P.S. Also, could you please resolve the conflicts?

assets/js/src/core/components/modal/alert-modal/hooks/use-alert-modal.tsx
sholzer: done :)

assets/js/src/core/components/modal/alert-modal/hooks/use-alert-modal.tsx
sholzer: done :)

assets/js/src/core/modules/app/error-handler/error-handler.tsx
sholzer: 1. done
2. The TypeScript check simply won't let me pass without the extra if for the type of the errorContent

ValeriaMaltseva (approved):

Issue #1170, opened 2025-03-24
[Simple search] Detail view


vin0401:
The different types need to be formatted and styled:

![Image](https://github.com/user-attachments/assets/2d4cf4ff-4f82-4cc8-8bd6-f3b9290d9e54)

markus-moser:
@ValeriaMaltseva Please discuss it with @vin0401. He prepared it already a bit.

ValeriaMaltseva:
Blocked by: https://github.com/pimcore/studio-backend-bundle/issues/955

PR #1208, opened 2025-03-27
[Simple search] Detail view

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/1170


vin0401 (changes requested):
Looks pretty good already.
But I would make the detail area also scrollable. 
It feels a bit strange that the modal is jumping around by just hovering over an Item. 
I think for now, we can simply go with a fixed height.

assets/js/src/core/modules/search/search.styles.tsx
```diff
@@ -0,0 +1,34 @@
...
+*  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+*  @license    https://github.com/pimcore/studio-ui-bundle/blob/1.x/LICENSE.md POCL and PCL
+*/
+
+import { createStyles } from 'antd-style'
+
+export const useStyles = createStyles(({ token, css }) => {
+  return {
+    detailContent: css`
+      max-height: 450px;
```
vin0401: I think it should be aligned with the results at the left.
Which are currently at 400px height.

PR #1291, opened 2025-04-02
Updated runner

bluvulture:
Resolves #


markus-moser:
@bluvulture Not sure how to proceed with this PR? Would like to either merge or close it.

bluvulture:
@markus-moser if you use php-stan I guess we then should merge it, but I see it has already been updated 

https://github.com/pimcore/studio-ui-bundle/blob/1.x/.github/workflows/static-analysis.yaml#L23

Issue #1135, opened 2025-03-14
Text file editiong lead to an error


drietsch:
Hello guys!

Found two topics in Pimcore Studio concerning the assets:

1.) Upload
If I upload a file, than I would expect, that the tree is reloading, which it does not.

2.) File Editing
If I want to edit a text file and publish it, a popup with a JSON error appears.

Cheers,
Dietz

markus-moser:
The upload is a separate topic which is still a bit work in progress so we handle it in separate issues. Let's tackle the 2.) with this one.

fashxp:
👍

PR #1302, opened 2025-04-03
[Assets] Add the Save option for the text type

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/1135


assets/js/src/core/modules/asset/draft/hooks/use-text-settings.ts
```diff
@@ -0,0 +1,124 @@
...
+    modifyDraft(state, action.payload.id, (draft: TextDataDraft): TextDataDraft => {
+      draft.textData = action.payload.textData ?? ''
+
+      markedAsModified(draft)
+
+      return draft
+    })
+  }
+
+  const removeTextData = (state: EntityState<TextDataDraft, number>, action: PayloadAction<TextDataAction>): void => {
```
markus-moser: Maybe we should remove the `addTextData`and `removeTextData` logic on all places? For me it seems a bit like a unused/unneeded feature which has quite a bit of code? `updateTextData` should be enough. WDYT?

markus-moser (changes requested):
Well done!

Please check my single comment and update it if my comment makes sense. Other then that we can merge it.

assets/js/src/core/modules/asset/draft/hooks/use-text-settings.ts
ValeriaMaltseva: @markus-moser Yep, let's do that! I was thinking about it before, but wasn't sure if we had any earlier plans related to this. I'll update it :)

markus-moser (approved):
Thanks a lot!

PR #1306, opened 2025-04-04
[Data Object Editor] Date & Time field should show time without seconds

### LINKED ISSUE #1301: [Data Object Editor] Date & Time field should show time without seconds
Like in the classic bundle we should only have hours and minutes:

Image

ValeriaMaltseva:
Resolves #1301


assets/js/src/core/components/date-picker/date-picker.tsx
```diff
@@ -32,6 +32,7 @@ export type DatePickerProps = PickerProps & {
   onChange?: (date: DatePickerValueType) => void
   outputType?: OutputType
   outputFormat?: string
+  showTime?: PickerProps['showTime']
```
markus-moser: Please check if this is needed or if it is already part of our extended props.

markus-moser (approved):
Please delete the commented line if possible. But in general it LGTM!

assets/js/src/core/components/date-picker/date-picker.tsx
ValeriaMaltseva: @markus-moser Thanks for the relevant comment; I've deleted it.

PR #1355, opened 2025-04-16
Button loading state changes size of the button

### LINKED ISSUE #1059: Button loading state changes size of the button
When a regular button with text is set to `loading: true` it loses the original size.

sholzer:
Resolves #1059


markus-moser (changes requested):
As discussed, the motion dev sometimes produces a unwanted padding in the button when no loading is involved.

markus-moser (approved):
Nice, now it works well!

Issue #1260, opened 2025-03-31
[Grid][Batch Edit] Close modal after apply changes


fashxp:
- apply changes with batch edit

![Image](https://github.com/user-attachments/assets/bb39b6a5-0d0c-4085-8fa9-40d62ae67c30)

- click `apply changes`

- changes get applied and batch edit modal remains open

![Image](https://github.com/user-attachments/assets/5d1a061d-c65a-4fd3-9cdd-72428dfd7408)

--> would close the modal and reload the grid

fashxp:
checked 👍

Issue #1143, opened 2025-03-17
[Batch Edit] Improvements


vin0401:
- Add missing edit types for assets
- The reload of the grid should be an manual action performed after an option is finished.
- Filter the drop down options to only editable fields and also only show options that are not already selected

markus-moser:
@xIrusux Please discuss with @vin0401 what we need to finalize the batch edit features, thanks

Issue #868, opened 2025-01-14
[Asset Batch Edit] Finalize Implementation


markus-moser:

markus-moser:
For a draggable window without background mask check this component:

https://github.com/pimcore/studio-ui-bundle/blob/1.x/assets/js/src/core/components/modal/window-modal/window-modal.tsx

PR #1359, opened 2025-04-16
[Batch Edit] Improvements

ValeriaMaltseva:
Tasks:
-  #868
-  #1143
-  #1260


assets/js/src/core/modules/data-object/listing/batch-actions/batch-edit-modal/batch-edit-modal.tsx
```diff
@@ -150,12 +156,36 @@ export const BatchEditModal = ({ batchEditModalOpen, setBatchEditModalOpen }: Ba
...
         // @todo change that to a more generic context
         assetContextId: id
       }))
     }
   }
 
+  const availableDropdownList = getAvailableColumnsDropdown(onColumnClick).menu.items
+
+  const getFilteredTypes = (column: any): object[] => {
+    return column?.children?.filter((child: any) => {
```
vin0401, line 172: I think we still miss here that we only show the fields that are really editable?
In the column should be a prop like editable.

vin0401 (changes requested):

vin0401 (approved):

PR #1366, opened 2025-04-17
Enter should trigger primary action in modal windows

### LINKED ISSUE #957: Enter should trigger primary action in modal windows
In modal windows (e.g. new folder), 'enter' should trigger the primary action (e.g. 'OK')

Image

sholzer:
Resolves #957


markus-moser (changes requested):
Thanks, in general it works well. But it seems that searching for the button and clicking it would have been the better idea (sorry for that!). The reason is the loading spinner of the button.

When you e.g. rename an asset in the asset tree the button gets a loading spinner while the item gets renamed. This loading spinner is not enabled when we press enter.

markus-moser:
Can we just always set the button to loading until its destroyed?

markus-moser (approved):

PR #1397, opened 2025-04-23
[Data Object][Grid] missing drop down options for filters

### LINKED ISSUE #1393: [Data Object][Grid] missing drop down options for filters
**Steps to reproduce**
- Login as admin (Studionightly)
- Open the 'Product Data' folder
- Select the car class
- Add the Objecttype as a filter field
- Open the dropdown

**Actual behaviour**
- The dropdown is empty (also happens on other fields)

Image

**Expected behaviour**
- Options for the Objecttype should be shown

**Environment**

- Platform Version Testsystem
- Windows
- Google Chrome

ValeriaMaltseva:
Resolves #1393


assets/js/src/core/modules/element/listing/decorators/general-filters/view-layer/components/sidebar/tabs/filters/field-filters/field-filters-container.tsx
```diff
@@ -20,20 +23,24 @@ import { FieldFilters, type FieldFiltersProps } from '@Pimcore/components/field-
 import { useFilter } from '../provider/filter-provider/use-filter'
 
 const FILTER_FIELD_KEY_IGNORE_LIST = ['size']
+const FILTER_TYPE_OVERRIDE_WHITELIST = ['dataobject.adapter', 'dataobject.objectbrick']
```
vin0401: I think it would be good to also move that into the dynamic types of the filters. 

Something like isOverridingFilterType or similiar.
With this we would have the benefit that we still can control the behavior via the plugins.

vin0401 (changes requested):

PR #1417, opened 2025-04-25
Add user permission checks to main nav

### LINKED ISSUE #1085: [Main Nav] Add permission checks for existing menu entries
Perspective permissions are already respected but we need to check also user permissions. We just need to set the `permission` property to the right permission string.

xIrusux:
Resolves #1085 

## Additional info
Tested user permission for: 
tags configuration, 
notes and events 
open document, 
open object, 
open asset,


xIrusux:
@markus-moser I could easily add an enum for those permissions but felt like there probably already is a place for them that I might have missed?

markus-moser:
@xIrusux We do not have one right now. On other locations where we somehow use the same permissions it is still hardcoded, e.g. here:
https://github.com/pimcore/studio-ui-bundle/blob/cbfd13f1a210830f7aee3d503940383408237609/assets/js/src/core/modules/element/editor/shared-tab-manager/tab-definitions.tsx#L70

So please add a new enum for this 👍

markus-moser:
@xIrusux Could you please add the missing checks for the user management + the enum?

markus-moser (changes requested):
Please check my comment ;)

PR #1439, opened 2025-05-06
hide asset and object widget if user does not have right permissions

### LINKED ISSUE #1429: [Permissions] - Hide Asset Tree and Object Tree when User does not have the corresponding Permissions

xIrusux:
Resolves #1429 

## Additional info
Hides asset / object widget if user does not have the correct permissions


markus-moser (approved):
Thanks, that's fine for now. I created a follow-up issue for a possible improvement. But this is not urgent at all:

https://github.com/pimcore/studio-ui-bundle/issues/1440

PR #1446, opened 2025-05-07
only show locale when there are localized fields available

### LINKED ISSUE #1371: [DataObjects][Bug] Locale switcher is displayed even if no localized fields are configured
The locale switcher in the dataobject is displayed even if there is no localizedfield present.
When you change the locale, it breaks the ui

xIrusux:
Resolves #1371 

## Additional info
- we are now hiding the locale switch when there are no localized fields available.


assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/localized-fields/localized-fields.tsx
```diff
@@ -22,7 +22,11 @@ export interface LocalizedFieldsProps extends AbstractObjectDataDefinition {
 }
 
 export const LocalizedFields = ({ children, noteditable, className }: LocalizedFieldsProps): React.JSX.Element => {
-  const { currentLanguage } = useLanguageSelection()
+  const { currentLanguage, setHasLocalizedFields } = useLanguageSelection()
+
+  useEffect(() => {
```
markus-moser, line 27: I'm not sure if this can lead to performance problems. Maybe it's better to do a simple `if (!hasLocalizedFields) { setHasLocalizedFields(true) }`?

assets/js/src/core/modules/data-object/editor/toolbar/context-menu/provider/layout-selection-provider.tsx
```diff
@@ -14,11 +14,15 @@ import { Content } from '@Pimcore/components/content/content'
 export interface ILayoutSelectionContext {
   currentLayout: string | null
   setCurrentLayout: (layoutId: string) => void
+  hasLocalizedFields: boolean
```
markus-moser: We have this logic twice - once in the layout-selection-provider and once in the language-selection-provider. Please remove it from the layout-selection-provider.

xIrusux: removed - thanks for the spot!

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/localized-fields/localized-fields.tsx
xIrusux: updated - put the additional check inside the useEffect that checks on first load

markus-moser (changes requested):
- Plese add translation keys for "On", "Between", "Before", "After"
- Currently filtering does not work correctly (often we send both from/to allthough only "After" is selected, sometimes it's not possible to apply the filter and trigger a new request)

markus-moser:
> * Plese add translation keys for "On", "Between", "Before", "After"
> * Currently filtering does not work correctly (often we send both from/to allthough only "After" is selected, sometimes it's not possible to apply the filter and trigger a new request)

Sorry, right message, wrong PR ;)

assets/js/src/core/modules/data-object/editor/toolbar/language-selection/language-selection.tsx
```diff
@@ -15,10 +15,15 @@ import { useLanguageSelection } from './provider/use-language-selection'
...
-  const { currentLanguage, setCurrentLanguage } = useLanguageSelection()
+  const { currentLanguage, setCurrentLanguage, hasLocalizedFields } = useLanguageSelection()
+
+  if (!hasLocalizedFields) {
+    return <></>
+  }
 
   return (
     <BaseLanguageSelection
+      key={ hasLocalizedFields ? 'localized' : 'not-localized' } // Force re-render when hasLocalizedFields changes
```
markus-moser: Hm... the `hasLocalizedFields` check cannot have any effect here as it's always true here (see if in line 20)

markus-moser (changes requested):
Please check my last comment, other then that it LGTM 👍

PR #1447, opened 2025-05-07
1042 relation data path formatter

### LINKED ISSUE #1042: [Data Object Editor] Relation data types path formatter

sholzer:
Resolves #1042


assets/js/src/core/modules/data-object/hooks/use-data-object-helper.ts
```diff
@@ -121,5 +128,37 @@ export const useDataObjectHelper = (): UseDataObjectReturn => {
...
+        id: item.id,
+        type: item.type,
+        label: item.fullPath,
+        path: item.fullPath,
+        nicePathKey: `object_${item.id}`
+      }
+      return acc
+    }, {})
+
+    const { data } = await store.dispatch(api.endpoints.dataObjectFormatPath.initiate({
```
markus-moser: Can we add some kind of caching here so that we don't request items with the same context information again after we already requested it? (so just remember the result in a local ref or so)

assets/js/src/core/modules/data-object/hooks/use-data-object-helper.ts
```diff
@@ -121,5 +128,37 @@ export const useDataObjectHelper = (): UseDataObjectReturn => {
...
+          containerType: 'object',
+          fieldname: fieldName,
+          objectId: dataObjectId,
+          layoutId: '0'
+        }
+      }
+    }))
+
+    if (data === undefined) {
+      throw new Error('Failed to format path: API response is undefined.')
```
markus-moser: Please add the `trackError(new ApiError(...))` logic here if RTK query returns an error in the request.

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/hooks/use-value.ts
```diff
@@ -33,22 +38,47 @@ interface UseValueReturn {
...
+        if (data.items !== undefined) {
+          const newValue = value.map((item) => ({
+            ...item,
+            fullPath: data.items.find(i => i.objectReference === `object_${item.id}`)?.formatedPath ?? item.fullPath
+          }))
+
+          setDisplayedValue(newValue)
+          setIsLoading(false)
+        }
+      }).catch(error => { trackError(new ApiError(error as unknown as ApiErrorData)) })
```
markus-moser: we should use `trackError(new ApiError())` only when we are sure that the error is an API error, therefore please move the logic to the other suggested location and do just a console.error here if we have other errors.

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/grid.tsx
```diff
@@ -94,7 +95,20 @@ export const ManyToManyRelationGrid = forwardRef(function ManyToManyRelationGrid
...
+          cell: (info) => {
+            return (
+              <Flex
+                align={ 'center' }
+                className={ 'p-mini' }
+              >
+                <Tag
+                  bordered={ false }
+                  color="geekblue"
+                ><SanitizeHtml html={ info.getValue() ?? '' } /></Tag>
```
markus-moser: This works for many-to-many-relations but unfortunately it seems that we need to find other locations where we need to apply this for the other multi-relation data types:

image

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/grid.tsx
```diff
@@ -94,7 +95,20 @@ export const ManyToManyRelationGrid = forwardRef(function ManyToManyRelationGrid
...
           },
-          size: 200
+          size: 200,
+          cell: (info) => {
+            return (
+              <Flex
+                align={ 'center' }
+                className={ 'p-mini' }
+              >
+                <Tag
```
markus-moser: If we have a path formatter involved we should not add the blue tag but just directly output the HTML as it is.

image

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/many-to-many-relation.tsx
```diff
@@ -79,7 +81,7 @@ export const ManyToManyRelation = (props: ManyToManyRelationProps): React.JSX.El
     }
   }, [JSON.stringify(props.value)])
 
-  if (props.isLoading === true) {
+  if (isLoading || props.isLoading === true) {
```
markus-moser, line 84: The loading state here seemed to be a good idea, but it's a bit annoying that the full data type has a loading state - especially when a new item is added via the tree. The grid data type itself also provides a loading prop. Maybe it's better to just set the grid itself to loading instead of everything? Not sure how easy it's possible but in a perfect world we would see the loading spinner directly on row level for newly added items (and only for the initial load we use the global one).

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/hooks/use-value.ts
```diff
@@ -33,22 +38,47 @@ interface UseValueReturn {
   addItems: (items: ManyToManyRelationValueItem[]) => void
   addAssets: (assets: Asset[]) => Promise<void>
   maxRemainingItems?: number
+  pathFormatterClass?: string | null
```
markus-moser, line 39: Currently we have the loading spinner + the triggered `format-path` request also when no path formatter is configured. We need to find a way to skip the path formatter logic when no `pathFormatterClass` is defined.

assets/js/src/core/modules/data-object/hooks/use-data-object-helper.ts
```diff
@@ -121,5 +128,37 @@ export const useDataObjectHelper = (): UseDataObjectReturn => {
...
+        nicePathKey: `object_${item.id}`
+      }
+      return acc
+    }, {})
+
+    const { data } = await store.dispatch(api.endpoints.dataObjectFormatPath.initiate({
+      body: {
+        objectId: dataObjectId,
+        targets,
+        context: {
```
markus-moser, line 156: It will be a bit tricky but we need to somehow find a way to respect the different possible context cases:

https://github.com/pimcore/studio-backend-bundle/blob/75ac539df3485d772261d8645d50765710ac0647/src/DataObject/Legacy/PathFormatterHelper.php#L46

markus-moser (changes requested):

PR #1459, opened 2025-05-08
1239 before and after filter

### LINKED ISSUE #1239: [Asset Grid][Filter] Add before and after filter
Image

In theory it could be done with between filter, but UX is clumsy to archive that. 
So either make that soother, or provide separate options

xIrusux:
Resolves #1239 

## Additional info
- add specific interface for setting "Before" and "After" date range filters

Screenshot from 2025-05-08 13-48-29


markus-moser (changes requested):
* Please add translation keys for "On", "Between", "Before", "After"
* Currently filtering does not work correctly (often we send both from/to allthough only "After" is selected, sometimes it's not possible to apply the filter and trigger a new request)

xIrusux:
@markus-moser 
After our conversation, I have applied form logic and also spent some time trying a different data format that works for both {on: dateValue} and  {from: dateValue, to:dateValue}. I went for a tuple of Unix timestamps and adapt it to the format required by the backend only on application.

This has removed a fair bit of code but there is another central question left regarding the final filter transformation and how to best handle the state change here. It would be great if we can have a conversation. 

Thanks

xIrusux:
@markus-moser 
I have now adjusted the code to:
- only have one state that changes and goes all the way to the server 
- state looks as follows:  
  interface FormValues {
    setting: DatePickerSettingValue
    from: number | null
    to: number | null
    on: number | null
  }
  
separated the "carry over date logic" into handleDatePickerValuesChange

the only detail that felt like a "nice to have" would be to also carry over the value when switching from BEFORE to AFTER
this requires thought cause the the "from and to" are tied and switching the settings a lot could make it confusing

xIrusux:
**Testing for carrying over data and applying the filter:**

legend:
-> = switch setting

on + apply = works
on -> after + apply = on value shows as after & apply works
after -> between + apply = after value shows as from & apply works
between -> change dates = apply works 
between -> before = to value shows as before & apply works 
on -> before = on value shows as before & apply works
before -> on = before value shows as on & apply works
before -> after = before value shows as after & apply works
after -> before = after value shows as before & apply works

markus-moser (approved):

PR #1495, opened 2025-05-15
252 notification list

### LINKED ISSUE #252: Notifications - Notification List
image

https://www.figma.com/design/pg65AmcoU5iVtQeroHK4iF/Full-Screens?node-id=11063-262383&t=jk20xqxzkpump3Pa-4

xIrusux:
Part of #252 

## Additional info

when clicking on "Notifications" inside the User menu ("Profile") -> open widget and gets notifications (basic setup)


xIrusux:
**status:** 
basic set up of the notification container (paging, toolbar, no content), api docs, notification api calls (hook), widget set up

**next:** 
- [x]  work on proper embedding of widget
- [x] improve hook
- [ ]  work on showing actual notification list

- [ ] api call to remove all notifications for user
- [ ] api call to remove specific notifications
- [ ] show attachment and make it clickable / allow opening

xIrusux:
@markus-moser 

also moved the main nav and user-menu into base-layout as discussed.

I have added a widget-definitions file as I thought it would be good to have a separate place to store those.
Please let me know if there is a better place as I could not find one that exists and fits the purpose. 

I have also shaped the widget similar to how they are shaped in the main nav:
i.e.:

export const NOTIFICATIONS: WidgetManagerTabConfig = {
  component: 'notifications',
  name: 'Notifications',
  id: 'notifications',
  config: {
    translationKey: 'notifications.label',
    icon: {
      type: 'name',
      value: 'notification'
    }
  }
}

I was unsure if some of this is unfitting i.e. the repeatet "noticiations" in the id 

the use-user-menu hook is still very much work in progress as not all calls are covered yet. This is just a first version

xIrusux:
@markus-moser put notification logic in its own folder together with the widget definition, hooks etc. 
Ready to review :)

assets/js/src/core/app/config/services/index.ts
```diff
@@ -142,7 +142,7 @@ import { DynamicTypeBatchEditElementDropzone } from '@Pimcore/modules/element/dy
 import { DynamicTypeObjectDataFieldCollection } from '@Pimcore/modules/element/dynamic-types/definitions/objects/data-related/types/dynamic-type-object-data-field-collection'
 import { DynamicTypeObjectDataObjectBrick } from '@Pimcore/modules/element/dynamic-types/definitions/objects/data-related/types/dynamic-type-object-data-object-brick'
 import { DynamicTypeGridCellDataObjectAdapter } from '@Pimcore/modules/element/dynamic-types/definitions/grid-cell/types/data-object-adapter/dynamic-type-grid-cell-data-object-adapter'
-import { MainNavRegistry } from '@Pimcore/modules/app/nav/services/main-nav-registry'
+import { MainNavRegistry } from '@Pimcore/modules/app/base-layout/nav/services/main-nav-registry'
```
markus-moser: Could you please rename the "nav" folder to "main-nav"?

assets/js/src/core/modules/notifications/hooks/use-notifications.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const useNotifications = (): UseNotificationsReturn => {
+  const [page, setPage] = useState<number>(1)
+  const [pageSize, setPageSize] = useState(20)
+  const [notifications, setNotifications] = useState<NotificationListItem[]>([])
+  const [totalItems, setTotalItems] = useState<number>(0)
+
+  const queryArgs: NotificationGetCollectionApiArg = useMemo(() => ({ body: { filters: { page, pageSize, includeDescendants: true } } }), [page, pageSize])
+
+  const [requestNotifications, { isLoading }] = useNotificationGetCollectionMutation()
```
markus-moser: Please add this endpoint to the query override list (https://github.com/pimcore/studio-ui-bundle/blob/1.x/assets/build/api/openapi-config.ts#L27) so that although this is a post request it's handled as query by RTK query. This will also change the usage/error handling here a bit.

For some endpoints we use `POST` as we could have quite complex filters which would not work for `GET` requests. But still they are queries and not mutations, therefore we should use this override feature.

assets/js/src/core/modules/notifications/hooks/use-notifications.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const getAllNotifications = useCallback(async (): Promise<void> => {
```
markus-moser: This `useNotifications` hook feels a bit like it is not optimal. It's a bit strange that `getAllNotifications` is a method which then sets a state and the state variables are returned directly from the hook. So when the `getAllNotifications` hook does not get called the other state variables are useless. Either this hook should be responsible for getting all notifications only (and then directly fetch and return them without additional method call) or `getAllNotifications` should work somehow isolated. I'd suggest to just use the hook for the listing and remove the `getAllNotifications` method.

markus-moser (changes requested):
In general it LGTM now but please check my comments.

assets/js/src/core/modules/notifications/hooks/use-notifications.tsx
xIrusux: thanks that makes so much sense. I felt it was very strange to have it as a mutation. / GET. Which threw me off a bit

markus-moser (approved):

PR #1528, opened 2025-05-23
252 notification detail

### LINKED ISSUE #252: Notifications - Notification List

xIrusux:
Resolves #252 

## Additional info

- [x] api calls to remove all notification for user
- [x] show attachment + click opens attachment
- [x] show as unread when unread (i.e. icon changes and goes green)
- [x] change icon as soon as opened i.e. read
- [x] ensure rtk query caches and reloads correctly
- [x] visually complete 
- [x] stop rtk query from triggering notification details call when notification isExpanded and deleted in that moment
- [ ] stop rtk query from reloading different page sizes on delete when user has previously selected more than one page sizing option 
- [ ] talk to backend about potentially sending a fullpath with the notification attachments so we dont have to get it via API call
- [x] raise backedn bug that paging does not work i.e. it always returns all notifications


xIrusux:
raised bug:

https://github.com/pimcore/studio-backend-bundle/issues/1105

xIrusux:
one problem persists:

- user selects different page sizes i.e. 19 / 20 / 50 
- user then deletes one notification 
- notifications rtk tag gets invalidated and may reload several page sizes for the query 

We have not really safe guarded against this in other similar paged situations. 
A way to do this would be to pass the pagesize into the rtk query. 

Let me know what you think. I can totally do this but thought if we decide to do it then it should be a pattern everywhere we have paging.

Thanks

assets/js/src/core/modules/notifications/notification-list.tsx
```diff
@@ -11,18 +11,26 @@
 import React from 'react'
 import { useNotifications } from './hooks/use-notifications'
 import { NotificationDetail } from './notification-detail'
+import { Space } from 'antd'
```
vin0401: Can we use our Space component here instead?

vin0401 (changes requested):

assets/js/src/core/modules/notifications/notification-list.tsx
xIrusux: uh thanks - changed :)

vin0401 (approved):

PR #1538, opened 2025-05-27
Notificaion polishing

xIrusux:
Resolves #

## Additional info

Polished Send Notification Modal:
- [x]  add loading state to send button (ie. custom modal footer)
- [x]  allow text search in select for notification recipient
- [x]  adjusted shared components to optionally allow the above
- [x]  show success snack bar message on successful sending of notification
- [x] add icon on send modal
- [x] show loading buffer when switching sides / on fetching not just loading

Adjust for backend updates:
- [x] ensure "hasAttachment" flag is used 
- [x] ensure paging now works
- [x]  fix faulty refetching of notifications in several page sizes when a notification gets deleted
- [x] combine and organize hooks in logical way
- [x]  raised issue regarding rework to get fullpath in notification detail query to remove un necessary api call to get attachment
https://github.com/pimcore/studio-backend-bundle/issues/1116

**2 Questions / Thoughts:**
- it feels a bit odd to me that one can only add one attachment
- ui of "attachment select" could be more intuitive.
As in below shot user hovers and sees the tooltip "Open" but icon is greyed out. Cause it is the first Icon it draws the users attention anyway. I think the other icons should be invisible
Screenshot from 2025-05-28 11-02-23


xIrusux:
[Screencast from 2025-05-28 12-43-29.webm](https://github.com/user-attachments/assets/8f684cb9-f93c-4f0a-8ba1-eaea9bc02653)

assets/js/src/core/modules/notifications/send-notification/send-notification-modal.tsx
```diff
@@ -44,22 +47,40 @@ export const SendNotificationModal = ({ open, ...props }: SendNotificationModalP
...
+        </Button>
+        <Button
+          loading={ isLoading }
+          onClick={ handleSend }
+          type='primary'
+        >
+          {t('user-menu.notification.send')}
+        </Button>
+      </ModalFooter>) }
       okText={ t('user-menu.notification.send') }
```
markus-moser: Let's remove the `okText` prop as with the custom footer this is obsolete

assets/js/src/core/modules/user/components/user-select/user-select.tsx
```diff
@@ -29,6 +29,7 @@ export const UserSelect = ({ onChange }: UserSelectProps): React.JSX.Element =>
       loading={ isLoading }
       onChange={ onChange }
```
markus-moser: We can remove this as it is also contained in the `...selectProps` right?

markus-moser (approved):
It LGTM - I just added 2 comments, please check them before merging it 👍

markus-moser:
> it feels a bit odd to me that one can only add one attachment

We cannot change this as we would need to also adjust it in the classic bundle. One attachment is fine for this use case.

> ui of "attachment select" could be more intuitive.
> As in below shot user hovers and sees the tooltip "Open" but icon is greyed out. Cause it is the first Icon it draws the users > attention anyway. I think the other icons should be invisible

I tend to agree but let's keep it like it is. At the moment in a lot of situations we do this and it's also done like this in the classic bundle. We would need to think about it in a holistic way - not that important for now...

PR #1550, opened 2025-06-02
[Grid] filtering for a specific date (on) does not work

### LINKED ISSUE #1514: [Grid] filtering for a specific date (on) does not work
**Steps to reproduce**

- Login as admin
- Open an assets folder
- Change the modification date of 3 objects
- Use the 'on' filter and search for the modification date

**Actual behaviour**
- No objects are shown (is the time also respected for on?)

Image

**Expected behaviour**
- Modified objects should be shown

**Environment**

- Playground
- Windows
- Google Chrome

ValeriaMaltseva:
Resolves #1514


assets/js/src/core/components/date-picker/utils/date-picker-utils.ts
```diff
@@ -29,9 +29,17 @@ export const fromDayJs = (value: Dayjs | null, outputType?: OutputType, outputFo
   if (value === null) {
     return null
   }
+
   if (outputType === 'timestamp') {
-    return value.unix()
+    const normalized = value.startOf('day')
```
markus-moser, line 34: Can't we just do value.startOf('day').unix() here?

assets/js/src/core/modules/element/dynamic-types/definitions/field-filters/components/dynamic-type-field-filter-datetime-component.tsx
```diff
@@ -90,7 +91,19 @@ export const DynamicTypeFieldFilterDatetimeComponent = (props: DynamicTypeFieldF
     }
   }
 
-  const handleDateChange = (field: 'on' | 'from' | 'to', value: number | null): void => {
+  const convertValueToISOFormat = (timestamp: number | null): string | null => {
+    if (timestamp === null) return null
+
+    return dayjs.unix(timestamp).format(DATE_FORMAT)
```
markus-moser: We should send a ISO date format with included timzone instead of just YYYY-MM-DD

markus-moser (changes requested):
Please check my comments. Maybe it could be a good idea to have a short discussion about it :)

assets/js/src/core/components/date-picker/utils/date-picker-utils.ts
ValeriaMaltseva: Unfortunately, no. 

**value.startOf('day').unix()** returns the date in UTC, but we need to show the date in local time.

markus-moser (approved):
LGTM

@heigpa Would be cool if you could test it.

heigpa:
works now. thx

PR #1580, opened 2025-06-04
1462 predefined properties

### LINKED ISSUE #1462: Studio UI Development - Predefined Properties
https://www.figma.com/design/pg65AmcoU5iVtQeroHK4iF/Full-Screens?node-id=11995-211968&t=mk2z52bVxwfdqj3Q-4

Image

xIrusux:
Resolves #1462


xIrusux:
Points for discussion:

- I have removed the modified Cells as it caused a fair amount of visual flicker - below the comparison

[Screencast from 2025-06-11 11-17-22.webm](https://github.com/user-attachments/assets/6b8c5476-b349-4478-8ec0-49e09aaf4b2c)

[Screencast from 2025-06-11 11-15-39.webm](https://github.com/user-attachments/assets/e26ac2f8-53dc-43b0-9c53-53c8bf73c647)

- User Permission tested 
- Nav Permissions - tried to test but it is no longer possible to change the perspectives for myself via admin?
(checked that the permission string is aligned with my enum)
- Added new enum folder for the select element tpes and content types. Not sure if this is somehow double up
- after every update / delete -> we do a refetch of all predefined properties. Is this ok?

- the value cell needs some work as it does not always function as it would make sense (already buggy inside custom metadata)
- there is also backend work needed to update the value. 
- this does raise the question whether I dont have enough info about how this should work. In the old UI it does not work atm.

markus-moser (approved):

PR #1613, opened 2025-06-17
Implement doc types

### LINKED ISSUE #1461: Studio UI Development - Document Types
https://www.figma.com/design/pg65AmcoU5iVtQeroHK4iF/Full-Screens?node-id=12066-558193&t=mk2z52bVxwfdqj3Q-4

Image

- [ ] Listing
- [ ] Adding New
- [ ] Deleting

xIrusux:
Resolves #1461

## Additional info
- search bar not possible as there are no filters allowed in get doc types query

[Screencast from 2025-06-18 15-23-52.webm](https://github.com/user-attachments/assets/bc7d08a1-d728-4bd2-9926-1319e9bd506c)


assets/js/src/core/modules/document-types/document-types-container.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const [documentTypeRows, setDocumentTypeRows] = useState<DocumentTypeRow[]>([])
+
+  const documentTypes = data?.items ?? []
+
+  const sortedRows = [...documentTypeRows].sort((a, b) => {
+    const dateA = a.creationDate ?? 0
+    const dateB = b.creationDate ?? 0
+    return dateB - dateA
+  })
```
Corepex, line 44: IMHO - that should be a backend task.

Corepex: IMHO - that should be a backend task.

assets/js/src/core/modules/document/document-api-slice-enhanced.ts
```diff
@@ -40,11 +56,18 @@ const api = baseApi.enhanceEndpoints({
 export type * from './document-api-slice.gen'
 
 export const {
-  useDocumentAddMutation,
-  useDocumentCloneMutation,
+  // useDocumentAddMutation,
+  // useDocumentCloneMutation,
```
Corepex: why do we need that?

Corepex (changes requested):
just some small questions

assets/js/src/core/modules/document-types/document-types-container.tsx
Corepex: discussed that point with @markus-moser and the conclusion was to sort the columns by `name`, not by `creationDate`.

Corepex:
I just saw one more thing. Please double-check the type options. I didn't see (for example) `hardlink` in the BE
image

assets/js/src/core/modules/document/document-api-slice-enhanced.ts
xIrusux: will check that - but there were already commented out lines when I worked on this file. Perhaps that has changed in master since. Will take a look

xIrusux:
@Corepex I am getting all the type options via api - if one is missing this would be a change that needs to happen in the backend

Corepex:
> @Corepex I am getting all the type options via api - if one is missing this would be a change that needs to happen in the backend

that's OK,but we should solve that problem or investigate why there is a difference - maybe the backend restructured something or its just a wrong result ...

Corepex (approved):
Besides the question with the types everything looks very good. GJ :1st_place_medal:  :astronaut:

xIrusux:
added small change - there was a property I could use to filter further. Thanks for the hint. 
There are still some differences from the types list in the new vs old api so I also pinged lukas to clarify. Thanks

PR #1616, opened 2025-06-18
[Data Object Editor] Sorting in multi relation fields

### LINKED ISSUE #1121: [Data Object Editor] Sorting in multi relation fields

ValeriaMaltseva:
Resolves #1121


assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/hooks/use-columns.tsx
```diff
@@ -0,0 +1,176 @@
...
+      size: 150
+    })
+  ]
+
+  const columns = !isUndefined(props.columnDefinition)
+    ? [...props.columnDefinition]
+    : defaultColumns
+
+  if (props.enableRowDrag) {
+    columns.unshift(
```
markus-moser, line 76: Wouldn't it be better to add e.g. a `enableRowDrag` prop into the grid component and add the column there? With this approach it would not be needed to expose `ROW_DRAG_COLUMN_ID`. By setting the prop everything would be handled automatically?

assets/js/src/core/components/grid/grid.tsx
```diff
@@ -322,20 +326,25 @@ export const Grid = ({
...
-                    isSelected={ row.getIsSelected() }
-                    key={ row.id }
-                    modifiedCells={ JSON.stringify(getModifiedRow(row.id)) }
-                    onFocusCell={ onFocusCell }
-                    onRowDoubleClick={ props.onRowDoubleClick }
-                    row={ row }
-                    tableElement={ tableElement }
-                  />
-                ))}
+                <SortableContext
```
markus-moser: The `SortableContext` provides a disabled property. We should disable it when drag and drop sorting is disabled. This would reduce the performance overhead and potential side effects.

assets/js/src/core/components/grid/grid-cell/grid-row.tsx
```diff
@@ -57,8 +86,9 @@ const GridRow = ({ row, isSelected, modifiedCells, ...props }: GridRowProps): Re
         row.getIsSelected() ? 'ant-table-row-selected' : '',
         props.onRowDoubleClick !== undefined ? 'hover' : ''
       ].join(' ') }
-      key={ row.id }
```
markus-moser, line 60: Not sure but isn't it a risky performance problem te remove the key here?

assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/components/many-to-many-relation/grid.tsx
```diff
@@ -225,20 +98,24 @@ export const ManyToManyRelationGrid = forwardRef(function ManyToManyRelationGrid
...
-            autoWidth
-            className={ props.className }
-            columns={ columns }
-            data={ getDataArray() }
-            disabled={ props.disabled === true || props.inherited === true }
-            onUpdateCellData={ props.onUpdateCellData }
-            resizable
-          />
+        <div style={ { maxWidth: 'calc(100% - 2px)' } }>
+          <DndContext
```
markus-moser: We should disable the DndContext when dnd sorting is not enabled. Also it would be nice if this is part of the grid component itself?

markus-moser (changes requested):
In general it works well. Please check my comments. :)

Also the drag and drop behaves sometimes strange when trying to drag outside of the grid. I can for example produce a situation where the data object toolbar is not at the bottom anymore:

image

And also a situation where I have a very long vertical scrollbar in the grid. It's hard to take a screenshot, but the steps are as following.

- Try to drag to another grid and keep dragging.
- Drag back to the original grid.
- Try to drag back at the bottom of the original grid

=> Depending on how I drag exactly, sometimes I get the massive scrollbar.

assets/js/src/core/components/grid/grid-cell/grid-row.tsx
ValeriaMaltseva: No, in our case it's fine because the **key** is already provided at the higher map level.

ValeriaMaltseva:
@markus-moser Thanks a lot for the comments! I've already updated the architectural approach.

Regarding this comment:
https://github.com/pimcore/studio-ui-bundle/pull/1616#pullrequestreview-2950412241

I cannot reproduce this behavior. Maybe the issue is gone because of the PR update, but I am not sure. Could you please check it again? If the issue still persists on your device, please try adding **autoScroll={false}** to the **DndContext**. It limits the user slightly but may help fix the scroll issues.

markus-moser:
@ValeriaMaltseva Thanks for updating the general approach - this LGTM now!

I still have scroll problems. The autoScoll={false} somehow improves it. But now I have somehow in general a doubled-scrollbar in my test data model:

image

When I scroll the outer scrollbar this happens:

image

When I go back to the 1.x branch the doubled scrollbar is gone. We can also take a look at it together at my local if this helps...

markus-moser (approved):

PR #1617, opened 2025-06-23
Fix the DynamicTypeFieldFilterSelectComponent

ValeriaMaltseva:
## Additional info
This PR fixes the bug related to using the Select type in the dynamic field type:
image
image


xIrusux:
Screenshot from 2025-06-25 10-34-19
 
Screenshot from 2025-06-25 10-39-22

 Above a screenshot of the select filter inside assets. 
 Here we rely on the options sitting inside config instead of config.fieldDefinition.options
 
We need to align the two different ways of handling this in assets vs data objects so the data is shaped the same way. 
Lets discuss if you need a second pair of eyes there :)

markus-moser (approved):

PR #1608, opened 2025-06-16
Implemented `Email Blocklist` 

Corepex:
## Changes in this pull request
Resolves #305

## Additional info
This pull request introduces functionality for managing an email blocklist in the application, including front-end components, API integration, and navigation updates. The changes add new tags for caching, enhance API endpoints, and implement UI components for viewing, adding, and removing blocked emails. Additionally, permissions and navigation items are updated to support the new feature.

### Email Blocklist Feature Implementation:

* **API Integration**:
  - Added new tags (`EMAIL_BLOCKLIST` and `EMAIL_BLOCKLIST_DETAIL`) for caching and invalidation in `emails-api-slice-enhanced`. Enhanced endpoints to provide and invalidate these tags for email blocklist operations.
  - Updated `tags.ts` to include new tags and their respective functions for providing and invalidating data related to the email blocklist. [[1]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013L47-R49) [[2]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013L96-R100) [[3]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013L138-R144)

* **UI Components**:
  - Introduced `EmailCard` component for displaying individual email blocklist entries with options to remove them.
  - Added `EmailBlocklistContainer` component for managing the blocklist, including pagination, refresh functionality, and adding new emails.
  - Implemented `use-email-blocklist` hook for handling blocklist operations, including adding and removing emails with error tracking.

### Navigation and Permissions Updates:

* **Navigation**:
  - Registered new navigation items for the email blocklist under "Tools/Email" in the main navigation, including widget configuration for the blocklist.
  - Updated imports in `bootstrap.ts` to include the new email module.

* **Permissions**:
  - Added a new `Emails` permission in `user-permission.ts` to manage access to the email blocklist feature.

PR #1660, opened 2025-07-08
Streamline email blocklist tab

Corepex:
Followup to #1608
Blocked by: https://github.com/pimcore/studio-backend-bundle/issues/1264

## Additional info
This pull request introduces improvements to the `EmailBlocklistContainer` component by enhancing state management and optimizing the loading behavior. The most important changes include adding `useEffect` to handle the `isFetching` state, modifying the query hook to include the `isFetching` property, and removing redundant state updates.

Enhancements to state management:

* [`assets/js/src/core/modules/email/blocklist/email-blocklist-container.tsx`](diffhunk://#diff-9bc044062a22be812730325d358b9b87fb5d9c4c60d17c69b1200019a576c5a8R47-R52): Added `useEffect` to monitor the `isFetching` property and update the `isLoading` state accordingly, ensuring the loading state is accurately managed.
* [`assets/js/src/core/modules/email/blocklist/email-blocklist-container.tsx`](diffhunk://#diff-9bc044062a22be812730325d358b9b87fb5d9c4c60d17c69b1200019a576c5a8L36-R36): Modified the `useEmailBlocklistGetCollectionQuery` hook to include the `isFetching` property, enabling finer control over the loading state.

Code cleanup:

* [`assets/js/src/core/modules/email/blocklist/email-blocklist-container.tsx`](diffhunk://#diff-9bc044062a22be812730325d358b9b87fb5d9c4c60d17c69b1200019a576c5a8L64-R70): Removed redundant `setIsLoading(false)` calls from the `onClick` handler, as the loading state is now managed by `useEffect`.

Dependency updates:

* [`assets/js/src/core/modules/email/blocklist/email-blocklist-container.tsx`](diffhunk://#diff-9bc044062a22be812730325d358b9b87fb5d9c4c60d17c69b1200019a576c5a8L22-R22): Added `useEffect` to the list of React imports to support the new loading behavior logic.


xIrusux (commented):
Just checked this out and tested it.

I  feel like it should not be possible to add the same email twice.
The backend does handle it and there is no duplicate shown after the action but I think it would be cleaner to not allow it in the FE.

It currently also does not refetch the api call when the window is closed and then re-opened. 
Which is implemented in send email for example and would be nice here too In case another agent has changed things. 

I tried the reloading behaviour if it was just dependant on isRTKLoading and it looked fine to me that way too. Thats more a heads up for next time. You dont  always need the double up loading state. 

Otherwise looks good to me.

idaiv:
Hi @Corepex! The design review is through as well, a couple of small details that need to be aligned:

- [x] Padding top and bottom of an item - 4 px and not 12 px as currently: 
image

You can see it here: https://www.figma.com/design/pg65AmcoU5iVtQeroHK4iF/Full-Screens?node-id=11063-262211&t=uVDR3Kb8j2tQllZV-4

- [ ] Email blocklist headline - the height is currently 48 px and it must be 44 px. 
- [x] Button - "New" and not "Add" - this is my mistake I forgot to update this instance of renaming. Sorry about it!
- [x] Blocklist - has its icon for the headline and for the tab - can you reproduce it in the open tab as well.

That's all, great work and thanks a lot!

Corepex:
> * [ ]  Email blocklist headline - the height is currently 48 px and it must be 44 px.

@idaiv i created a followup for the blocklist headline since this has to be solved everywhere.

https://github.com/pimcore/studio-ui-bundle/issues/2112

PR #1738, opened 2025-07-18
Reports View - Grid & Drill-down Filters

### LINKED ISSUE #854: Reports View - Grid & Drill-down Filters
Image

- [x] Grid with paging

Image

- [x] Column configuration in sidebar

Image

- [x] Drill-Down Filters

Image

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=8306-237176&t=tkonv6NnTEp682z2-0

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/854


assets/js/src/core/modules/reports/reports-view/context/grid-context.tsx
```diff
@@ -0,0 +1,70 @@
...
+  drillDownFilters?: Record<string, any>
+}
+
+interface IGridContext {
+  columns: Array<AccessorKeyColumnDef<unknown, any>>
+  setColumns: (columns: Array<AccessorKeyColumnDef<unknown, any>>) => void
+  filters?: IFilterValue
+  setFilters: (filters: IFilterValue) => void
+  resetFilters: () => void
+}
```
vin0401: I'm thinking if it maybe makes sense to split those up in two different contexts - column and filter context
Otherwise the filters would rerender when you change the columns and vice versa due to the behavior of react context.

I'm also thinking if we can abstract that in general a bit further. 
So that this is not unique to reports, and also reusable for other areas like dependencies and so on.

We could even think of splitting the filters even further between drill down and column filters.

vin0401 (changes requested):

vin0401 (approved):

Issue #1676, opened 2025-07-09
[Documents][Context Menu Options][Detail view] Open in new Window


markus-moser:
<img width="362" height="328" alt="Image" src="https://github.com/user-attachments/assets/c8061ed1-9a2b-4cc6-aa9d-946e0f6848df" />

Same as: https://github.com/pimcore/studio-ui-bundle/issues/1672

In the new UI its the more menu at the bottom toolbar:

<img width="351" height="490" alt="Image" src="https://github.com/user-attachments/assets/1577e55b-5dfb-4629-91d2-6843e4f34598" />

Issue #1672, opened 2025-07-09
[Documents][Context Menu Options][Tree] Open in new Window


markus-moser:
<img width="416" height="635" alt="Image" src="https://github.com/user-attachments/assets/ff036637-8906-45e8-8582-55030ba204df" />

<img width="1413" height="676" alt="Image" src="https://github.com/user-attachments/assets/e9fd3ed1-c4ea-44d5-ab43-31cb45b06503" />

PR #1767, opened 2025-07-25
[Documents][Context Menu Options][Tree] Open in new Window #1672

MSchinwaldEl:
Implements the context menu entry for "open in new window"
- in tree (#1672)
- in bottom toolbar (#1676)

## Additional info
The basic functionality of the context menu entry is implemented. When the context menu item is clicked, a request to the Document API is fired and a window with the value of settingsData.url is opened.

**I have a few questions though:**

### Hardlinks
I checked the behavior in the classic ui bundle (https://github.com/pimcore/admin-ui-classic-bundle/blob/7bf40f83bc1e93bd58decc2db98a85015a5d8461/public/js/pimcore/document/tree.js#L635) and saw that it should also work for hardlinks. Currently I think the Document API doesn't return the correct URL that we need to open the window. 

Here's an example of the response:
```json
{
    "fullPath": "/en/More-Stuff/Developers-Corner/Hardlink/Magazine",
    "published": true,
    "type": "hardlink",
    "key": "Magazine",
    "index": 1,
    "hasChildren": false,
    "hasWorkflowWithPermissions": false,
    "permissions": {
        "save": true,
        "unpublish": true,
        "list": true,
        "view": true,
        "publish": true,
        "delete": true,
        "rename": true,
        "create": true,
        "settings": true,
        "versions": true,
        "properties": true
    },
    "isSite": false,
    "navigationExclude": false,
    "id": 122,
    "parentId": 121,
    "path": "/en/More-Stuff/Developers-Corner/Hardlink/",
    "icon": {
        "type": "name",
        "value": "hardlink"
    },
    "userModification": 1566913143,
    "creationDate": 1566912993,
    "modificationDate": 2,
    "userOwner": 2,
    "locked": null,
    "isLocked": false,
    "additionalAttributes": [],
    "customAttributes": {
        "icon": null,
        "tooltip": null,
        "additionalIcons": [],
        "key": null,
        "additionalCssClasses": []
    },
    "editableData": null,
    "missingRequiredEditable": false,
    "settingsData": {
        "sourceId": 46,
        "propertiesFromSource": false,
        "childrenFromSource": true,
        "sourcePath": "/en/Magazine"
    },
    "draftData": null,
    "hasWorkflowAvailable": false
}
``` 
Do we need to request the Document API using settingsData.sourceId? I'm not very familiar with the hardlink feature so I'm not sure what's the best approach here.

### Permissions
What I also noticed in the classic ui bundle is that the code checks perspectiveCfg.inTreeContextMenu("document.open"). Do we need this check too? How would I implement it?
I also saw that in some places studio ui bundle `isTreeActionAllowed` is used, but there's currently no TreePermission for _open_. Do we need this check?


fashxp:
@MSchinwaldEl hi and welcome on board :heart_eyes: 
Can you please sign the CLA by pasting the 'I have..' sentence as comment (see comment above). Thx

MSchinwaldEl:

markus-moser:
@MSchinwaldEl 
Regarding your questions:
- Hardlinks: Please just disable the feature for hardlinks for now. I checked it also in the classic bundle and there it does not seem to work too. Let's just not provide it for hardlinks.

- Permissions: The mentioned helpers and enums are the correct ones. We need to add the "open" Permission there. You will find a list of all permissions in this request: `/pimcore-studio/api/perspectives/configuration/studio_default_perspective`.  For our tree context menu actions the contextPermissions of the document tree are the relevant ones:

image

Please check the failed pipeline. It seems that there is something where our typescript type checker is not happy about it.

Other then that it LGTM

markus-moser (changes requested):
Please check my comment

MSchinwaldEl:
@markus-moser
Thanks for the review!
I made the following changes:

- Hardlinks: We don't provide the functionality for them for now.
- Permissions: Added the enums and check the treepermission for open.
- I fixed the TS error: There seems to be a type missing in the API definition. For now I added a type guard. I added an issue in the backend bundle: https://github.com/pimcore/studio-backend-bundle/issues/1263

assets/js/src/core/modules/document/actions/open-in-new-window/use-open-in-new-window.tsx
```diff
@@ -0,0 +1,97 @@
...
+      hidden: isContextMenuEntryHidden(document),
+      onClick: async () => {
+        await openInNewWindow(document.id, onFinish)
+      }
+    }
+  }
+
+  const openInNewWindowTreeContextMenuItem = (node: TreeNodeProps): ItemType => {
+    return {
+      label: t('document.open-in-new-window'),
```
markus-moser, line 83: Please add a translation to our studio.en.yaml

assets/js/src/core/modules/document/actions/open-in-new-window/use-open-in-new-window.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const dispatch = useAppDispatch()
+  const { isTreeActionAllowed } = useTreePermission()
+
+  const openInNewWindow = async (
+    documentId: number,
+    onFinish?: () => void
+  ): Promise<void> => {
+    setIsLoading(true)
+    const { data } = await dispatch(api.endpoints.documentGetById.initiate({ id: documentId }))
```
markus-moser: Please add a `trackError(new ApiError(error))` when the fetch returns an error.

markus-moser (changes requested):
@MSchinwaldEl Thanks a lot, in general it looks good. Please check my 2 comments, then we can merge it!

MSchinwaldEl:
> @markus-moser Thanks for the review! I made the following changes:
> 
> * Hardlinks: We don't provide the functionality for them for now.
> * Permissions: Added the enums and check the treepermission for open.
> * I fixed the TS error: There seems to be a type missing in the API definition. For now I added a type guard. I added an issue in the backend bundle: [[Documents][Open in new Window] Specify type of settingsData more detailed studio-backend-bundle#1263](https://github.com/pimcore/studio-backend-bundle/issues/1263)

Regarding the type of settingsData, I will just leave the type guard for now, since it can't be properly typed because it's a dynamic array (see https://github.com/pimcore/studio-backend-bundle/issues/1263)

markus-moser:
> Regarding the type of settingsData, I will just leave the type guard for now, since it can't be properly typed because it's a dynamic array (see [pimcore/studio-backend-bundle#1263](https://github.com/pimcore/studio-backend-bundle/issues/1263))

Can we define a type on client side instead? So something like this?

```typescript
type SettingsData = {
   url?: string
}

const settingsData: SettingsData | undefined = data?.settingsData ?? undefined
const url = settingsData?.url
```

We can trust the backend that the URL will be either a string or undefined and this would clean-up our code :)

Also added a hint how we could improve the type guard approach using lodash. Also this one would be fine, just would like to get rid of the complex type checking logic.

assets/js/src/core/modules/document/actions/open-in-new-window/use-open-in-new-window.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const { data, error } = await dispatch(api.endpoints.documentGetById.initiate({ id: documentId }))
+
+    if (!isUndefined(error)) {
+      trackError(new ApiError(error))
+      setIsLoading(false)
+    }
+
+    // type guard to check if settingsData has a url property, since the settingsData object in the API response can vary
+    function hasUrl (obj: unknown): obj is { url: string } {
+      return typeof obj === 'object' && obj !== null && 'url' in obj && typeof (obj as any).url === 'string'
```
markus-moser: If we want to keep the type guard we should use lodash as it could simplify the logic a lot

```typescript
isPlainObject(obj) && isString(obj?.url)
```

MSchinwaldEl: I simplified it with
`has(data?.settingsData, 'url') && isString(data?.settingsData.url)`
and got rid of the `hasUrl` function

markus-moser (approved):
Now it looks good, let's merge it 👍

PR #1808, opened 2025-08-01
[Assets] Layout image preview not correct

### LINKED ISSUE #1742: [Assets] Layout image preview not correct
**Steps to reproduce**

- Login as admin
- Open an image folder
- Check the layout

**Actual behaviour**
- Layout not correct

Image

**Expected behaviour**
- Layout should be correct

**Environment**

- Studio Nightly
- Windows
- Google Chrome

ValeriaMaltseva:
Resolves #1742


markus-moser (changes requested):
The images are now somehow stretched to 100% width and height. This can change there aspect ratio. We need to do them max-width and max-height 100% instead and then center it within the preview teaser.

markus-moser (changes requested):
A small detail. The images/icons should also be vertically centered:

Image

PR #1818, opened 2025-08-04
[Documents][Context Menu Options][Tree] Add document

### LINKED ISSUE #1671: [Documents][Context Menu Options][Tree] Add document
Image

MSchinwaldEl:
Implementation of [Documents][Context Menu Options][Tree] Add document #1671


markus-moser:
@MSchinwaldEl What does it need to finish this PR?

assets/js/src/core/modules/document/tree/context-menu/context-menu.tsx
```diff
@@ -36,6 +38,34 @@ export const DocumentTreeContextMenu = (props: DocumentTreeContextMenuProps): Re
...
+    modalTitle: t('document.tree.context-menu.add-snippet')
+  })
+  const { addDocumentTreeContextMenuItem: addEmailTreeContextMenuItem } = useAddDocument({
+    type: 'email',
+    iconValue: 'mail-02',
+    contextMenuKey: ContextMenuActionName.addEmail,
+    formType: AddDocumentFormType.KEY_ONLY,
+    modalTitle: t('document.tree.context-menu.add-email')
+  })
+  const { addDocumentTreeContextMenuItem: addNewsletterTreeContextMenuItem } = useAddDocument({
```
markus-moser: Newsletter is part of an external bundle so we need to remove it here and add it later via the bundle (https://github.com/pimcore/newsletter-bundle)

assets/js/src/core/modules/document/actions/add-page/use-add-document.tsx
```diff
@@ -0,0 +1,320 @@
...
+        </Form.Item>
+      </Form>
+    )
+  }
+
+  // Modal form content component for document creation
+  const ModalFormContent: React.FC<{ form: FormInstance<any>, firstInputRef: React.RefObject<InputRef>, t: any }> = ({ form, firstInputRef, t }) => {
+    return formType === AddDocumentFormType.FULL
+      ? (
+        <FullFormContent
```
markus-moser: For both modal variants:

- Do we have a chance to auto-focus the first input when the modals get opend?
- Can we auto submit the form when pressing enter?

We integrated the same already in our form modal hook -> see https://github.com/pimcore/studio-ui-bundle/pull/1366. The confirm of useFormModal would be exactly what we need for the single modal here... so if it is easer we can also use the confirm for this instead of your custom implementation. Would also be ok when we have the auto-submit feature only for the single field modal but not for the modal with 3 fields. But the focus would really be also an advantage for the 3-field variant.

assets/js/src/core/modules/document/actions/add-page/use-add-document.tsx
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    modal.confirm({
+      icon: null,
+      title: modalTitle,
+      content: <ModalFormContent
+        firstInputRef={ firstInputRef }
+        form={ form }
+        t={ t }
```
markus-moser: we can also use `useTranslation` within the ModalFormContent instead of passing the t function ;)

assets/js/src/core/modules/document/actions/add-page/use-add-document.tsx
```diff
@@ -0,0 +1,320 @@
...
+        language: null,
+        translationsSourceId: null,
+        inheritanceSourceId: null
+      }
+    })
+
+    try {
+      const response = await createDocumentTask
+
+      if (response.error !== undefined) {
```
markus-moser: Very minor, but just as a hint, we prefer using lodash for those type checks. We would then have `!isUndefined(response.error)`

markus-moser (changes requested):

assets/js/src/core/modules/document/actions/add-page/use-add-document.tsx
MSchinwaldEl: - I implemented the auto-focus for both modals
- Submission on enter is implemented for the modal with the single input (by using useFormModal) but for now I would rather not implement it for the modal with multiple inputs. I already invested a lot of time and there's always something that just doesn't work out. Maybe we can put it in a separate issue for later.

markus-moser (approved):

PR #1777, opened 2025-07-28
Implemented `Redirects` tab

xIrusux:
## Changes in this pull request
Resolves #1645 (Part 1)

This pull request introduces support for managing SEO redirects in the application. It adds a new redirects module, updates permissions and navigation, enhances the icon library with new icons, and improves the grid cell component to better handle string values and text input. The most important changes are grouped below.

## Redirects Module Integration

* Added the new redirects module, including import and registration in the application bootstrap (`@Pimcore/modules/redirects`), and set up API configuration for SEO redirects endpoints. [[1]](diffhunk://#diff-252d7c8a4651b914260fd7855ebe378b9274fbef63a80faebd45493ed4baa499R30) [[2]](diffhunk://#diff-a24dac5d3f3852f17d3d694a6e9a8e306da8b84516e7a40bc92f17f457b65876R40-R43) [[3]](diffhunk://#diff-a24dac5d3f3852f17d3d694a6e9a8e306da8b84516e7a40bc92f17f457b65876R170-R172)
* Introduced the `BeginnerRedirectModal` component for creating new redirects with a simple form interface.
* Added CSV import modal styles for the redirects module.

## Permissions and Navigation

* Added new `Redirects` entries to user permissions and navigation permissions enums to control access to the redirects feature. [[1]](diffhunk://#diff-caa0e3cdaa3221365d5b799c2a7749e83a11f85339ea3aed81cabe9f92eab3e2R26) [[2]](diffhunk://#diff-a02820fd7bf44eda873ddb5e0ecd73a64a416649a721b8c6dc9286134bf4d04eR17)

## Tag System Enhancements

* Added `REDIRECTS` to the tag names and updated `providingTags` and `invalidatingTags` to support cache invalidation and data fetching for redirects. [[1]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013R46) [[2]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013R98) [[3]](diffhunk://#diff-b4cbe56cce73117505d8014cd7b9d020f41f740a0790f643e494fd445c261013R153)

## Icon Library Updates

[...]

PR #1850, opened 2025-08-07
Add required messages to create redirect modal + remove html entities when reverting to plain text

xIrusux:
Followup to https://github.com/pimcore/studio-ui-bundle/pull/1777

## Additional info
Small fixes
- [x] add required messages on redirect modal
- [x] remove html entities when revertig back from translation html to plain text


Corepex (approved):
codewise everything looks good

PR #1865, opened 2025-08-08
[Documents][Context Menu Options][Detail view] Open preview in new window

### LINKED ISSUE #1677: [Documents][Context Menu Options][Detail view] Open preview in new window
Image

Very similar to https://github.com/pimcore/studio-ui-bundle/issues/1676

MSchinwaldEl:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/1677


markus-moser (changes requested):
There are some details which we need to change:

1. The regular open should open the URL but the preview is a bit different. It needs to open the document **within the same domain** with the pimcore_preview paramter. Otherwise the session cookie will not be valid. Therefore we need to build the URL based on the full path of the document.

2. In the detail view we need to show the open preview in new window feature (and also the open) also for snippets etc. Take a look at the old UI and compare for which document types it is available.

markus-moser (changes requested):
For me both "open in new window" and "open preview in new window" do the same e.g. for ducment ID 136.

This doesn't seem to be fullfilled, at least not for pages:

> The regular open should open the URL but the preview is a bit different. It needs to open the document within the same domain with the pimcore_preview paramter. Otherwise the session cookie will not be valid. Therefore we need to build the URL based on the full path of the document.

Can you also please change it so that we send "pimcore_preview=true&pimcore_studio_preview=true"` (e.g. `http://localhost/de/shared/teasers/Magazin?pimcore_preview=true&pimcore_studio_preview=true&_dc=1756806089063`)? We added an additional check to the core which handles the preview slightly different when the additional parameter is sent...

MSchinwaldEl:
> For me both "open in new window" and "open preview in new window" do the same e.g. for ducment ID 136.
> 
> This doesn't seem to be fullfilled, at least not for pages:
> 
> > The regular open should open the URL but the preview is a bit different. It needs to open the document within the same domain with the pimcore_preview paramter. Otherwise the session cookie will not be valid. Therefore we need to build the URL based on the full path of the document.
> 
> Can you also please change it so that we send "pimcore_preview=true&pimcore_studio_preview=true"`(e.g.`[http://localhost/de/shared/teasers/Magazin?pimcore_preview=true&pimcore_studio_preview=true&_dc=1756806089063`](http://localhost/de/shared/teasers/Magazin?pimcore_preview=true&pimcore_studio_preview=true&_dc=1756806089063%60))? We added an additional check to the core which handles the preview slightly different when the additional parameter is sent...

1. Seems like I accidentally removed the `!` in the check for whether it's the preview. I have now fixed it.
2. I added the `pimcore_studio_preview` param.

markus-moser (approved):
Now it LGTM 👍

PR #1983, opened 2025-08-31
Pull Request Proposal: Export RTK Query Hooks in Studio UI Bundle SDK

drietsch:
## Problem Statement

When developing Pimcore Studio extensions that need to access class definitions and other Pimcore data, developers cannot use the RTK Query hooks that are already available internally in the Studio UI Bundle. These hooks are generated from the OpenAPI specification but are not exposed through the SDK's Module Federation exports.

This leads to developers having to either:
1. Use mock data instead of real Pimcore data
2. Make direct API calls instead of using the existing RTK Query infrastructure
3. Duplicate the RTK Query setup in their own bundles

## Solution

Export the RTK Query hooks and types through the SDK's main.ts file to make them available to extension developers via Module Federation.

## Changes Made

`/vendor/pimcore/studio-ui-bundle/assets/js/src/sdk/main.ts`

```typescript
// Export RTK Query hooks for class definitions
export { 
  useClassDefinitionCollectionQuery,
  useClassDefinitionGetQuery,
  useClassCustomLayoutCollectionQuery,
  usePimcoreStudioApiClassCustomLayoutGetQuery
} from '@Pimcore/modules/class-definition/class-definition-slice.gen'

// Export types for class definitions
export type {
  ClassDefinitionListItem,
  ClassDefinitionCollectionApiResponse,
  ClassDefinitionCollectionApiArg
} from '@Pimcore/modules/class-definition/class-definition-slice.gen'
```

## Benefits

1. **Developer Experience**: Extension developers can use the same RTK Query hooks that Pimcore Studio uses internally
2. **Consistency**: Ensures extensions use the same data fetching patterns as the core application
3. **Type Safety**: Provides proper TypeScript types for API responses
4. **Performance**: Leverages RTK Query's built-in caching and optimization
5. **Maintenance**: Reduces code duplication between core and extensions

## Example Usage

After this change, extension developers can import and use these hooks:

```typescript
import { 
  useClassDefinitionCollectionQuery,
  useClassDefinitionGetQuery
} from '@pimcore/studio-ui-bundle'

export const MyComponent = () => {
  // Get all class definitions
  const { data: classData, isLoading } = useClassDefinitionCollectionQuery()
  
  // Get specific class schema
  const { data: schema } = useClassDefinitionGetQuery(
    { dataObjectClass: 'Product' },
    { skip: !classData }
  )
  
  // Use real Pimcore data in the component
  return (
    <div>
      {classData?.items.map(cls => (
        <div key={cls.id}>{cls.name}</div>
      ))}
    </div>
  )
}
```

## Testing

1. Built the Studio UI Bundle with the new exports
2. Used the exported hooks in a real extension (TiptapSyntaxEditor component)
3. Verified that real Pimcore class data is accessible
4. Confirmed no breaking changes to existing functionality

## Next Steps

This pattern should be extended to other RTK Query slices as well:
- Asset management hooks
- Data object hooks
- Tag management hooks
- User and permission hooks

## Migration Guide

No breaking changes. This is purely additive - existing code continues to work unchanged.

## Related Issues

- Module Federation cannot access internal modules that aren't explicitly exported
- Extension developers need access to the same data fetching infrastructure as core

## Request for Feedback

This is a proposal for improving the SDK's Module Federation exports. Please review and let me know if:
1. There are any concerns about exposing these hooks
2. Additional hooks should be included
3. The export structure should be organized differently


markus-moser:
Hi @drietsch,

Thanks for your PR, but there are some things which we need to check:

1. I cannot see any changes in main.ts in this PR -> did you select the wrong source branch?
2. We export the api RTK query hooks already via separate exports, see:
https://github.com/pimcore/studio-ui-bundle/tree/c90a9389da2f8efa9a789d51fda39db00ecbc3c4/assets/js/src/sdk/api

At the moment it seems that the class-definitions hooks are really missing there, so we need to add them there instead of the main export. You would then be able to import it like this:

```typescript
import { 
  useClassDefinitionCollectionQuery,
  useClassDefinitionGetQuery
} from '@pimcore/studio-ui-bundle/api/class-definition'
```

vin0401:
@drietsch - We just added these to the sdk. 
All class definition related hooks/types are now available over the `@pimcore/studio-ui-bundle/api/class-definition` import, like @markus-moser has proposed. 

Closing this in favour of https://github.com/pimcore/studio-ui-bundle/pull/1993

PR #1992, opened 2025-09-01
[Usermanagement] unable to open user detail page #1961

### LINKED ISSUE #1961: [Usermanagement] unable to open user detail page
**Steps to reproduce**

- Login as admin
- Open the detailpage of a user

**Actual behaviour**
- Error message appears

Image

**Expected behaviour**
- Detailpage should be opened

**Environment**

- Studio Nightly
- Windows
- Google Chrome

sholzer:
Resolves #1961


markus-moser:
@sholzer Could you please check the linter error in the frontend build pipeline?

markus-moser:
Hi @sholzer,
I found out that the problem is that we now have in the availableAdminLanguages only the locale string anymore. We have a different approach for the name lookup now, please check:
https://github.com/pimcore/studio-ui-bundle/pull/1992/commits/f1cd98686b07648e3a29518bf4408b6dd194948d

Could you please adjust the other locations in the user management and profile to the new approach? I found out that e.g. the language selects are relevant too...

Issue #1756, opened 2025-07-24
Studio UI Development - Custom Reports Editor - SQL Adapter


fashxp:
<img width="832" height="862" alt="Image" src="https://github.com/user-attachments/assets/aae3c896-4baf-4173-89ef-469387bee8b9" />

ValeriaMaltseva:
Blocked by: https://github.com/pimcore/studio-backend-bundle/issues/1345

ValeriaMaltseva:
Blocked by: https://github.com/pimcore/studio-backend-bundle/issues/1354

PR #2043, opened 2025-09-11
Studio UI Development - Custom Reports Editor - SQL Adapter

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/1756


assets/js/src/core/modules/reports/dynamic-types/definitions/definition-adapters/dynamic-type-definition-abstract.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+import { type ReactElement } from 'react'
+import { injectable } from 'inversify'
+import { type DynamicTypeAbstract } from '@Pimcore/modules/element/dynamic-types/registry/dynamic-type-registry-abstract'
+import { type IReportConfigurationSectionProps } from '@Pimcore/modules/reports/reports-editor/types'
+
+@injectable()
+export abstract class DynamicTypeDefinitionAbstract implements DynamicTypeAbstract {
```
markus-moser: This should contain something like `CustomReport` or so in the name.

assets/js/src/core/modules/reports/dynamic-types/definitions/definition-adapters/dynamic-type-definition-registry.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import { injectable } from 'inversify'
+import { DynamicTypeRegistryAbstract } from '@Pimcore/modules/element/dynamic-types/registry/dynamic-type-registry-abstract'
+import {
+  type DynamicTypeDefinitionAbstract
+} from '@Pimcore/modules/reports/dynamic-types/definitions/definition-adapters/dynamic-type-definition-abstract'
+
+@injectable()
+export class DynamicTypeDefinitionRegistry extends DynamicTypeRegistryAbstract<DynamicTypeDefinitionAbstract> {}
```
markus-moser: This should contain something like `CustomReport` or so in the name.

assets/js/src/core/modules/reports/dynamic-types/definitions/custom-report-definition-adapters/dynamic-type-custom-report-definition-abstract.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { injectable } from 'inversify'
+import { type DynamicTypeAbstract } from '@Pimcore/modules/element/dynamic-types/registry/dynamic-type-registry-abstract'
+import { type IReportConfigurationSectionProps } from '@Pimcore/modules/reports/reports-editor/types'
+
+@injectable()
+export abstract class DynamicTypeDefinitionAbstract implements DynamicTypeAbstract {
+  abstract readonly id: string
+  abstract readonly label: string
+
+  abstract getElement ({ currentData, updateFormData }: IReportConfigurationSectionProps): ReactElement
```
markus-moser: I wonder if we should give this a bit of a more specific name? We never know if we need to at some point to extend the interface here for other purposes and then just `Element` is a bit of a random name.

markus-moser (changes requested):
Only a few hints regarding namings. Didn't test it, but in general it looks good!

Also compared it with the version of the classic bundle and it seems that there also the configuration form part needed to be flexible (https://github.com/pimcore/pimcore/blob/12.x/bundles/CustomReportsBundle/public/js/pimcore/report/custom/definitions/sql.js). So I'd say we are fine!

Didn't test it, but for me it's fine to merge it!

PR #2054, opened 2025-09-12
Fix user password saving

### LINKED ISSUE #1604: [User] saving a new user does not work
**Steps to reproduce**

- Login as admin
- Create a new user
- Provide a name and a password
- Save and publish

**Actual behaviour**

- Nothing happens

Image

**Expected behaviour**

- User with password should be saved

**Environment**

- Studio Nightly
- Windows
- Google Chrome

sholzer:
Resolves #1604


markus-moser:
We need to wait for this before we can fully finalize it:
https://github.com/pimcore/studio-backend-bundle/issues/1362

markus-moser:
I'll merge it so that user saving works. The other detail will be solved after the backend API is adjusted.

PR #2057, opened 2025-09-12
[Data Object] - Fix and allow "published" field to be updated when in Grid

### LINKED ISSUE #1959: [DataObject] - updating for published boolean does not work
Image

Image

xIrusux:
Resolves #1959 

## Additional info
This pull request refactors how updates are sent to the API when editing data objects inline, specifically handling updates to the `published` column differently from other columns. The main change is the introduction of a conditional structure that customizes the payload for published status updates to ensure correct API behavior.

**Inline Edit API Update Logic:**

* In `use-inline-edit-api-update.tsx`, added a conditional to check if the updated column is `published`. If so, the payload now includes the `published` value at the root level and sends an empty `editableData` object; otherwise, the payload continues to update the appropriate field inside `editableData`.


assets/js/src/core/modules/data-object/listing/decorator/inline-editing/hooks/use-inline-edit-api-update.tsx
```diff
@@ -63,16 +63,26 @@ export const useInlineEditApiUpdate = (): UseInlineEditApiUpdateReturn => {
...
-          {
-            id: update.id,
+    const isPublishedColumn = columnKey === 'published'
+
+    const dataItem = {
+      id: update.id,
+      ...(isPublishedColumn
+        ? {
+            published: value,
+            editableData: {}
```
markus-moser: Please remove this line as the editableData is allowed to be undefined.

markus-moser (changes requested):
Only a small hint, but other then that it LGTM 👍

Issue #1745, opened 2025-07-22
[Relations] values not shown


heigpa:
*Steps to reproduce**

- Login as admin
- Open an accessory
- Check Compatible to

**Actual behaviour**
- No values are shown

<img width="1455" height="767" alt="Image" src="https://github.com/user-attachments/assets/14cbb0de-bcb3-49cb-97a6-dd18673d7789" />

**Expected behaviour**
- Values should be shown

**Environment**

- Studio Nightly
- Windows
- Google Chrome

markus-moser:
This is the same as this:
https://github.com/pimcore/studio-ui-bundle/issues/899

heigpa:
works now. thx

PR #2070, opened 2025-09-17
[Data Object Editor] Integrate visible columns data

### LINKED ISSUE #899: [Data Object Editor] Integrate visible columns data

ValeriaMaltseva:
Resolves #899 #1745


markus-moser:
See also https://github.com/pimcore/studio-backend-bundle/issues/1389

@ValeriaMaltseva Please prepare as discussed that we use the `props.combinedFieldName` instead of the regular field name to request the visible columns from the backend (so that it also works if the field is used within a container data type like objectbricks). Hopefully this is the only FE change regarding this backend task. But in addition it would be cool if you could check the 2 cases from the backend task as soon as the backend task is resolved. So maybe we should create a follow-up ticket for testing this?

markus-moser:
Just as a reference - we need to change the way how the grid column type is handled.

https://github.com/pimcore/studio-ui-bundle/blob/c47a15b167d14bbb6a38c74977c8180487387d3e/assets/js/src/core/modules/data-object/listing/listing-container.tsx#L65

https://github.com/pimcore/studio-ui-bundle/blob/c47a15b167d14bbb6a38c74977c8180487387d3e/assets/js/src/core/modules/element/listing/abstract/view-layer/components/grid/hooks/use-grid-options.tsx#L33

markus-moser:
This happens within the relation visible columns (tested it for a reverse many to many relation field). Doesn't happen in the regular grid. It seems that theere is some "base styling" different:

https://github.com/pimcore/studio-ui-bundle/issues/2093

markus-moser:
We should also add caching if possible to avoid that the same element needs to be refetched again when adding new items.

ValeriaMaltseva:
> This happens within the relation visible columns (tested it for a reverse many to many relation field). Doesn't happen in the regular grid. It seems that theere is some "base styling" different:
> 
> #2093

I'll fix this bug in a separate PR.

ValeriaMaltseva:
@markus-moser I've checked all the comments and updated the PR. Could you please review it once again?

markus-moser:
@ValeriaMaltseva Will do it - we have 4 open sonar issues -> could you please either accept or solve them?

markus-moser (changes requested):
@ValeriaMaltseva Re-tested everything and in general it looks really good, just found 2 topics:

- The approach with getting the column definition out of the dynamic types works for many to many object relation and advanced many to many object relation, but not for reverse many to many relation

- Created this follow-up ticket which is not urgent but should also be solved at some point:
https://github.com/pimcore/studio-ui-bundle/issues/2113

ValeriaMaltseva:
@markus-moser I've created a follow-up ticket: https://github.com/pimcore/studio-ui-bundle/issues/2119

PR #2122, opened 2025-09-25
System info for various elements

### LINKED ISSUE #1513: System info for various elements
Will there be an overview of the system infos for the different elements (data objects, assets, documents) like in the classic-ui?

Image

ValeriaMaltseva:
Resolves #1513


assets/js/src/core/modules/element/hooks/use-element-actions-menu.tsx
```diff
@@ -0,0 +1,128 @@
+/**
```
markus-moser, line 1: Can you please move the `element-toolbar` out of the components folder somewhere into `modules/element` (+ remove the story) and then also move this hook there? It's not something which should be in storybook and therefore it's also nothing which should be in `core/components`. 

As you can see we have it in "refactor" - and maybe now it's a good time to do so? 

Image

assets/js/src/core/modules/element/hooks/use-element-actions-menu.tsx
```diff
@@ -0,0 +1,128 @@
...
+import type { DropdownMenuProps } from '@Pimcore/components/dropdown/dropdown'
+import { Flex } from '@Pimcore/components/flex/flex'
+import { Text } from '@Pimcore/components/text/text'
+import { Icon } from '@Pimcore/components/icon/icon'
+import { type Element, getElementDeeplink } from '@Pimcore/modules/element/element-helper'
+import type { ElementType } from '@Pimcore/types/enums/element/element-type'
+import { useSystemInfoModal } from '@Pimcore/modules/element/components/system-info-modal/hooks/use-system-info-modal'
+
+export type IElement = Element & {
+  fileSize: number
```
markus-moser, line 23: Should be `fileSize` and `mimeType` optional here? We only have it for assets...

assets/js/src/core/modules/app/modals-provider.tsx
```diff
@@ -36,7 +37,9 @@ export const ModalsProvider = ({ children }: ModalsProviderProps): React.JSX.Ele
               <VideoModalProvider>
                 <SendTestEmailProvider>
                   <SiteModalProvider>
-                    {children}
+                    <SystemInfoModalProvider>
```
markus-moser: In general we should try to not add all modals to the modals provider because at some point we will have a massive amount of nested modal providers then. Initially I created the modal provider to provide a way to access certain modals from the document editor iframe. For sure for other locations we also have sometimes the requirement to add a modal to a central location in the DOM. But there would be also 2 alternative ways to achieve this:

1. Create an "app modal" like described in the ant design docs: https://ant.design/components/app. We do this in a lot of situations and the advantage is that you can open a modal with an imperative call instead of adding a modal element to the DOM (this would be probably the best for this case as it is only a modal which shows some information)

2. I also lately added a generic `ModalHolder` provider:

https://github.com/pimcore/studio-ui-bundle/blob/112f2a3f6e36501609959b6c8ecc429bd25ae52f/assets/js/src/core/modules/document/actions/translations/use-translations.tsx#L42

https://github.com/pimcore/studio-ui-bundle/blob/112f2a3f6e36501609959b6c8ecc429bd25ae52f/assets/js/src/core/modules/app/modals-provider.tsx#L31

This can be used to inject simple modals without specific requirements. 

For me its also fine when we don't refactor this case, but wanted to mention the possibilities so that you are aware of it ;)

markus-moser (changes requested):
Just some comments, but in general it LGTM 👍

PR #2155, opened 2025-09-30
[Translations] - adjust to type

### LINKED ISSUE #1971: [Translation] - adjust updating of type functionality after backend updated
this currently works but is a bit of a hack. 
It will be clean once backend implemented this:
https://github.com/pimcore/studio-backend-bundle/issues/1285

but will need a adjustments of the lines

xIrusux:
Resolves #1971 

## Additional info
This pull request simplifies the logic for updating translations in the `useTranslation` hook by removing special handling for the `'type'` column and streamlining how translation data is constructed and sent to the API.

**Refactoring and simplification:**

* Removed the branch that handled the `'type'` column separately, so now translation updates are handled consistently regardless of the column.
* Simplified the construction of the `translationData` array, reducing code duplication and making the update logic more maintainable.


Corepex (approved):
LGTM

PR #2203, opened 2025-10-06
[Data Object Editor] Search should also search in visible columns data

### LINKED ISSUE #2113: [Data Object Editor] Search should also search in visible columns data
When I search for "Congress" in this example it should still display the single entry:

Image

ValeriaMaltseva:
Resolves #2113


assets/js/src/core/components/many-to-many-relation/utils/helpers.ts
```diff
@@ -29,3 +30,15 @@ export const getElementCellConfig = (disabled?: boolean): ElementCellConfig => {
     }
   }
 }
+
+export const flattenValues = (value: unknown): string[] => {
+  if (isNil(value)) return []
+
+  if (isString(value)) return [value]
```
markus-moser: Can we improve this so that we also check for numbers here? So that if the original value is a number that we also find it in the search. But only if the full value is a number, not of it is an object and one of it's attributes is a number (so not for the `flatMap` version of flattenValues)...

markus-moser (changes requested):
Just one thing, the rest LGTM :)

Issue #2212, opened 2025-10-07
[Data Object Grid][Advanced columns] Add missing transformers


markus-moser:
Need to add:
- [x] twig operator
- [x] anonymizer
- [x] element counter
- [x] string replace
- [x] substring
- [x] translate
- [x] trim
- [x] php code
- [x] boolean formatter
- [x] dateformatter
- [x] explode
- [x] blur.

Additionally, need to add the transformers grouping inside the dropdown.

markus-moser:
Please discuss with @vin0401 what needs to be done.

PR #2322, opened 2025-10-10
[Data Object Grid][Advanced columns] Add improvements

ValeriaMaltseva:
Task: https://github.com/pimcore/studio-ui-bundle/issues/2212


assets/js/src/core/components/pipeline/item/dynamic-group/dynamic-group-dropdown.tsx
```diff
@@ -32,15 +34,21 @@ export const DynamicGroupDropdown = ({ children, dynamicTypeRegistryId }: Dynami
...
-    label: t(`grid.advanced-column.advancedColumns.${dynamicType.id}`),
-    onClick: () => {
-      operations.add({
-        key: dynamicType.id
-      })
-    }
-  }))
+  const { transformersMenuItems } = useTransformersMenuItems(availableDynamicTypes)
+
+  const isTransformersRegistry = dynamicTypeRegistryId === serviceIds['DynamicTypes/Grid/TransformersRegistry']
```
vin0401: I would rather implement this in the DynamicTypePipelineAbstract Class.
Because now this only works for transformers and it's also not really adjustable for the community.

So for example we could add a group property for all the dynamic types that defaults to null. 
When the group is null, we simply show the item on the first level of the drop down.
When the group is set, we just group it based on the matching groups over all dynTypes.
May we even go for an string array, so that we can support nesting of multiple levels. :) 

This way, we just have to set the group once in the dynType, when we introduce one for example via Bundle. :)

vin0401 (changes requested):

vin0401 (approved):

PR #2427, opened 2025-10-28
[Notifications] Add SDK exports

sholzer:


assets/js/src/sdk/modules/notifications/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
markus-moser, line 1: @sholzer We also need to add the new file to rsbuild.sdk.config.ts, otherwise it will not work

markus-moser (changes requested):

assets/js/src/sdk/modules/notifications/index.ts
sholzer: done :)

PR #2470, opened 2025-10-31
Update the global context logic

ValeriaMaltseva:
Task: https://github.com/pimcore/copilot-bundle/issues/691


assets/js/src/core/modules/asset/editor/editor-container.tsx
```diff
@@ -38,7 +39,7 @@ const EditorContainer = (props: EditorContainerProps): React.JSX.Element => {
 
   useEffect(() => {
     if (isWidgetActive) {
-      setContext({ id })
+      setContext({ id, context: getBaseAssetContext(asset) })
```
markus-moser: It feels a bit strange that we have a "context" in the `context` - maybe we should name this somehow different - e.g. "contextIdentifiers" or similar? then we should also name those `getBaseAssetContext` somehow with the same naming convention...

markus-moser (changes requested):
As discussed, we should maybe rethink the naming convention a bit, other than that we can do it like that.

PR #2534, opened 2025-11-13
[Docs] Revise installation commands for Pimcore Studio UI

fashxp:
Updated installation instructions for the Pimcore Studio UI npm package to include React version compatibility.


doc/04_Plugins_Getting_Started.md
```diff
@@ -16,15 +16,17 @@ To get started, create a assets root folder for your application (e.g. `/assets`
 [Pimcore Studio UI npm package](https://www.npmjs.com/package/@pimcore/studio-ui-bundle?activeTab=readme), which offers full TypeScript support for the Pimcore SDK.
 
 ```
-npm install @pimcore/studio-ui-bundle
+npm install @pimcore/studio-ui-bundle react@18.3.x react-dom@18.3.x
```
mattviasat, line 19: No-blocking, but is there any way this could be added to platform-version? It could be something like a file containing the version. Or a CLI tool that just printed the correct version to install. 
I'm sure there is some way.
This is low priority, but it would be nice if everything was defined through platform-version.

fashxp: yes, we are already looking for ways to force that somehow... that is also the reason this PR is not merged yet :)

fashxp:
@vin0401 already any findings on that one?

vin0401:
@fashxp - Nothing really final, yet.

But most likely we can solve it via peerDependencies in the package.json.
So that any bundle that requires the package will automatically get the right version or a conflict, when another version is  already installed.

But it is also not the highest priority right now.

fashxp:
let's merge this now, and create a followup-ticket.

PR #2726, opened 2025-12-17
Feature/add configurable custom report delimiter

torqdev:
We previously submitted a feature request to allow the delimiter to be edited on the fly when exporting custom report data via csv in the legacy admin ui: https://github.com/pimcore/pimcore/pull/18691.

We added the logic into StudioUI to allow for the same configuration.


Corepex (approved):

Corepex:
@torqdev thanks for making studio even more awesome :1st_place_medal: :rocket:

PR #2795, opened 2026-01-19
Add prefix for custom color tokens

markus-moser:
This is a follow-up to https://github.com/pimcore/studio-ui-bundle/pull/2776

We need to introduce a prefix for the additional color tokens as otherwise we got conflicts with the ant design default colors.

image


idaiv:
Updated on Figma with the prefix colorCoding (token example colorCodingGold2).
Synced with the Icon Library so that the icons color application is updated too.
Available in the [Ant Design 5.11 file](https://www.figma.com/design/lomefPHRhJnkljB2nXtuKV/Ant-Design-5.11?node-id=516-40776&p=f&t=hMI7QwaSQ2QZWcZw-0) - both in the Variables and in Theme Buddy.

PR #2951, opened 2026-02-15
Fix translations edit modal

fashxp:
translations edit modal is broken right now: 
- tab changes are not possible
- entering translations is not possible
- restoring to plain text is not possible

image

this PR fixes it, and also adds some data-testids


fashxp:
friendly :ping_pong: @xIrusux ;-)

xIrusux:
Hi @fashxp  

Thanks for finding the bug and this approach to fixing it. 

While testing the changes in this PR, I noticed that what is displayed in the plain text tab switches quite uncontrollably between html an plain which can be seen below. 

I also wanted to avoid the  isRestoring + setTimeout approach which binds to Reacts render scheduling and is not guaranteed to work every time.

The root cause seems to have been that showOnlyHtmlTab was derived from the content shape rather than explicit intent.

Testing the changes in this PR: 
[Screencast from 2026-03-09 16-03-21.webm](https://github.com/user-attachments/assets/0e2aa672-7745-4c53-9d43-a7730020fb95)
 
Here my follow-up PR
https://github.com/pimcore/studio-ui-bundle/pull/3055

After changes in above PR:
[Screencast from 2026-03-09 16-16-22.webm](https://github.com/user-attachments/assets/d3e40b62-9017-4716-a8c2-b697ad6aa244)

I think the video shows how this was intended to function. I.e. the "plain text" tab disappears entirely once one has switched to "html"

Let me know what you think :) 
Thanks!

fashxp:
@xIrusux thx for checking and improving ;-) ... could also add the data-testids to #3055, then we can get rid of this one, right?

xIrusux:
closing this PR as discussed - https://github.com/pimcore/studio-ui-bundle/pull/3055 was merged

PR #2952, opened 2026-02-16
More data testids

fashxp:
added more data-testids 
also optimized the modals by [capturing the form instance at creation time](https://github.com/pimcore/studio-ui-bundle/commit/5997e7ff3e44f1e6148536aba78e7b6d2aee4ed7)


assets/js/src/core/components/modal/form-modal/hooks/use-form-modal.tsx
```diff
@@ -128,6 +135,14 @@ export function withInput (props: InputFormModalProps, onKeyBoardSubmit, onSetMo
     ...modalProps,
     type: props.type ?? 'confirm',
     icon: props.icon ?? null,
+    okButtonProps: {
```
markus-moser: maybe a bit picky, but we have this fragment with okButtonProps and cancelButtonProps now on 4 different locations in this file, would be nice to have a simple helper function for this to avoid duplicate code here...

markus-moser (approved):
Please check my single comment. In general it's fine to merge it, but would be a nice 2 have to get rid of those duplicated lines...

assets/js/src/core/components/modal/form-modal/hooks/use-form-modal.tsx
fashxp: done

markus-moser (approved):

PR #2988, opened 2026-02-20
Class Editor | Add new dynamic types for layout components: Accordion, Region, Field…

chilladelia:


chilladelia:
@markus-moser i already prepared the next PR. We need to discuss how we implement the text component or if we leave it for the moment:
image

Also: is wysiwyg already available to be integrated in that location?

markus-moser (approved):

PR #3006, opened 2026-02-26
[Task] Use 2026.1

martineiber:


composer.json
```diff
@@ -18,7 +18,7 @@
   "minimum-stability": "dev",
   "require": {
 	"php": "~8.3.0 || ~8.4.0",
-    "pimcore/pimcore": "^12.0",
+    "pimcore/pimcore": "^12.0 || ^2026.1",
```
markus-moser, line 21: it LGTM but I wonder if we also need a branch alias similar to this?
https://github.com/pimcore/pimcore/blob/2026.x/composer.json#L207

martineiber: We can add it like we have it in Studio BE. 
https://github.com/pimcore/studio-backend-bundle/blob/1.x/composer.json#L66

markus-moser: Lets skip the branch alias for now and add it as soon as we need it

PR #3018, opened 2026-03-02
Fixed asset download button

### LINKED ISSUE #2889: [GDPR] image export does not work
Search for an image in the GDPR extractor and export the image.
The export does not work.

Image

Corepex:
Resolves #2889 

## Additional info
This pull request refactors the `ExportButton` used in the assets tab of the GDPR data extractor module. The main changes involve moving the `ExportButton` component to a new local directory, updating its implementation, and adjusting its usage to provide a more accurate filename for exports.

Component refactoring and relocation:

* The `ExportButton` component is now imported from a new local path (`./components/export-button/export-button`) instead of a shared location, indicating it has been moved and possibly specialized for the assets tab.

Export button implementation improvements:

* A new `ExportButton` implementation was added, which constructs the export URL, dynamically sets the download filename based on the asset's path, and triggers a download of the exported asset as a `.zip` file.
* The `ExportButton` now receives a `filename` prop (derived from the asset's path) and no longer receives the `providerKey` prop, simplifying its interface and making the filename more descriptive for the user.


assets/js/src/core/modules/gdpr-data-extractor/components/tab-panel/components/assets-tab/assets-tab.tsx
```diff
@@ -87,8 +87,8 @@ export const AssetsTab = ({ data, providerKey, onSortingChange, ...props }: Asse
           <Flex>
             <ExportButton
               data-testid={ `gdpr-assets-export-${data.id}` }
+              filename={ data.fullPath.split('/').pop() ?? crypto.randomUUID() }
```
xIrusux: if fullpaPath is malformed or ends with / the user could end up downloading something like: 9f6b7a-...zip maybe more human meaningful would be to mix the data.id with something like `asset-${data.id}`
```suggestion
              filename={ data.fullPath.split('/').pop() || `asset-${data.id}` }
```

assets/js/src/core/modules/gdpr-data-extractor/components/tab-panel/components/assets-tab/components/export-button/export-button.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { getPrefix } from '@Pimcore/app/api/pimcore/route'
+
+interface ExportButtonProps extends Omit<React.ComponentProps<typeof IconButton>, 'id' | 'icon'> {
+  id: number
+  filename: string
+}
+
+export const ExportButton = ({ id, filename, onClick, ...iconButtonProps }: ExportButtonProps): React.JSX.Element => {
+  const handleExport = (e: React.MouseEvent<HTMLElement, MouseEvent>): void => {
+    const sanitizedFilename = filename.replaceAll(/\s/g, '_')
```
xIrusux: I am also just using sanetization a lot. To make it more robust you can also sanetize characters that are problematic on some systems not just white spaces 

```suggestion
   const sanitizedFilename = filename
  .trim()
  .replace(/\s+/g, '_')
  .replace(/[\\/:*?"<>|]/g, '_')
```

xIrusux (commented):
some small suggestions but otherwise LGTM! :)

assets/js/src/core/modules/gdpr-data-extractor/components/tab-panel/components/assets-tab/assets-tab.tsx
Corepex: ur right - I added it as a fallback if something fails - but `asset-${data.id}` is definitely more user-friendly than a random string. Very good point. :100:

assets/js/src/core/modules/gdpr-data-extractor/components/tab-panel/components/assets-tab/components/export-button/export-button.tsx
Corepex: also a good point. I merged the two replaces into one expression.

PR #3019, opened 2026-03-02
Set page to 1 on tab change

### LINKED ISSUE #2884: [GDPR] reset active page after switching element
**Steps to reproduce**
- Login as admin
- Open the GDPR extractor
- Search for vw
- Switch page size to 10 on data objects
- Navigate to the last page
- Switch to assets

**Actual behaviour**
- Nothing is shown as the active page is not reset

Image

**Expected behaviour**
- When switching to other elements (data objects, assets, sent mails...) the active page always should be reset

**Environment**

- Studio Nightly
- Windows
- Google Chrome

Corepex:
Resolves #2884


xIrusux (approved):
LGTM :)

PR #3263, opened 2026-04-01
fix: check permissions before restoring static widgets to prevent 403 errors

### LINKED ISSUE #3262: [Notifications] - Error when permission revoked
To reproduce: 

1. Log in as a user with notifications permission and open the notifications panel
2. Revoke the notifications permission from that user
3. Log in again — an error modal appears as the notifications widget tries to restore

a user who previously had the notifications panel open loses the notifications permission. On their next login, the widget manager restores the notifications widget from persisted state and fires API calls that return 403

xIrusux:
## Summary

- Adds an optional `permission` field to `WidgetManagerTabConfig` so static widgets can declare a required `UserPermission`
- `StaticWidgetRestorer.restore()` now checks `isAllowed()` before restoring a widget, silently skipping widgets the user lacks permission for
- The notifications widget now declares `permission: UserPermission.Notifications`, preventing 403 errors when a user without that permission logs in and their previous session had the notifications panel open

## Problem

When a user without the `notifications` permission logged in (or had their permission revoked), the `StaticWidgetRestorer` would unconditionally restore the notifications widget from the persisted layout state. This caused the notifications API calls to fire, resulting in 403 errors displayed as error modals.

Unlike `MainNavWidgetRestorer` (which already checks `isAllowed()` and `isAllowedInPerspective()`), the `StaticWidgetRestorer` had no permission checking at all.

## Changes

| File | Change |
|------|--------|
| `widget-manager-slice.ts` | Added optional `permission?: UserPermission` to `WidgetManagerTabConfig` interface |
| `static-widget-restorer.ts` | Added `isAllowed()` check in `restore()` — returns `false` if user lacks the declared permission |
| `notifications/index.tsx` | Set `permission: UserPermission.Notifications` on the `NOTIFICATIONS` config |

## Related

- Closes https://github.com/pimcore/studio-ui-bundle/issues/3262
- Backend companion PR: https://github.com/pimcore/studio-backend-bundle/pull/1741 (removes `IsGranted` from admin settings GET endpoint to fix the other 403-on-login issue)


Corepex (approved):
LGTM

PR #3282, opened 2026-04-02
[Login] Skip unauthenticated API requests on login page

### LINKED ISSUE #3249: [Loginpage] Multiple failed requests when loading login page
Image

currently on the nightly I see 
- /pimcore-studio/api/translations/create
- /pimcore-studio/api/user/current-user-information
- /pimcore-studio/api/bundle/translations-provider-interfaces/config/ui-settings
- /pimcore-studio/api/translations/create
- /pimcore-studio/api/user/current-user-information
- /pimcore-studio/api/bundle/translations-provider-interfaces/config/ui-settings

they all fail with 401 ... 

I guess they never should be fired on the login page...

xIrusux:
## Summary

Resolves #3249 

- Adds an authentication guard to the missing translations middleware so it does not fire `POST /translations/create` when the user is not logged in
- On the login page, i18n encounters unknown keys and the debounced listener dispatches the request after 3 seconds — this fails with 401 since no session exists. The guard short-circuits when `state.authentication.isAuthenticated !== true`.

## Context

When loading the login page, several API requests fail with 401:

- `/pimcore-studio/api/translations/create` — **fixed here**: the missing translations middleware now checks auth state before dispatching
- `/pimcore-studio/api/user/current-user-information` — **by design**: this is the auth probe that determines whether to show the login page
- `/pimcore-studio/api/bundle/translations-provider-interfaces/config/ui-settings` — originates from `translations-provider-interfaces` bundle, needs a separate fix there


Corepex (approved):
LGTM

PR #3287, opened 2026-04-07
Fix reset color sending null instead of empty string to backend

### LINKED ISSUE #3267: [Appearance and Branding] - resetting the colors does not work
When resetting the colors the following error message appears when saving,

Image

xIrusux:
## Summary

- Fixes the `formatColor` function in `color-panel.tsx` to return an empty string (`''`) instead of `null` when a color is cleared via the ColorPicker's "allow clear" action.
- The backend `Branding` schema declares `brandColor` and `backgroundShade` as non-nullable `string` types, so sending `null` caused the validation error: *"This value should be of type string."*
- Empty string `''` is the established sentinel for "no color" throughout the codebase (used in form `initialValues`, the backend hydrator, and the TypeScript type definition).

Resolves #3267


Corepex (approved):
LGTM

PR #3294, opened 2026-04-07
Fixed elementType missmatch in search

### LINKED ISSUE #3177: [GlobalSearch][Data Objects] Data Object Not Openable From “All” Tab
In Studio UI, data objects cannot be opened from the global search when using the “All” tab. Selecting a data object from the “All” results does not trigger any action. The same object can be opened without issues when using the “Data Objects” tab. This issue prevents one of our **enterprise customers** from effectively managing their data as they rely on Studio UI for several reasons.

**Steps to Reproduce**
- Open global search
- Stay in the “All” tab
- Search for a data object
- Click on object to open it

**Actual Behavior**
No action is triggered when clicking a data object in the “All” tab.

**Expected Behavior**
Clicking a data object in the “All” tab should open the object, consistent with the behavior in the “Data Objects” tab.

**Affected Versions**
0.14.9 (our version)

**Environment**
- OS: macOS 26.3
- OS: Windows 11
- Browser: Google Chrome Version 144.0.7559.133 (Official Build) (arm64)

Corepex:
Resolves #3177 

## Additional info
This pull request refactors the way the element type is handled when opening a search result item. Instead of directly using the `elementType` from the search result, it now maps the value using a helper function to ensure the type is valid before proceeding.

**Element type handling improvements:**

* Replaced direct use of `item.elementType` with a call to `mapToElementType`, and added a check to prevent opening elements with an unmapped or invalid type.
* Removed the import of the `ElementType` enum, as it is no longer directly used in the component.
* Updated the hook usage to destructure `mapToElementType` from `useElementHelper`.


xIrusux (approved):
LGTM :)

PR #3313, opened 2026-04-08
Fixed language flag issue

### LINKED ISSUE #3261: [Object Editor] - missing country flag
I added Italien (it_IT) as a new language in the system settings and when checking the language in the editor, the flag is missing. 
For some other countries (e.g. Austria) it works. I can't tell how many languages are missing the country flag.

Image

Corepex:
Resolves #3261 

## Additional info
This pull request makes a small adjustment to the logic for resolving which flag icon to display for a given language code. The change simplifies the conditional check for displaying a country flag by removing an unnecessary comparison.

- Simplified the condition in `resolveLanguageFlag` within `flag-icon.tsx` by removing the redundant `countryCode !== languageOnly` check, ensuring that as long as `countryCode` exists and the flag exists, it will be used.


xIrusux (approved):
LGTM :)

PR #3321, opened 2026-04-08
Added `priority` translation to detail modal

### LINKED ISSUE #3157: [Application Logger] Detail Popup -> for type use name instead of number
7 should be `info` - as in the grid.

Image

Corepex:
Resolves #3157 

## Additional info
This pull request makes a small change to the `DetailModal` component to improve the display of the priority field. The priority value is now translated using the appropriate translation key instead of displaying the raw value. 

- The `priority` field in the `formData` object is now set to the translated string using the `t` function and the `application-logger.filter.priority-level` translation namespace, ensuring user-friendly display of priority levels.


xIrusux (approved):
LGMT :)

PR #3328, opened 2026-04-10
Added `onFinish` callback to config create

### LINKED ISSUE #2973: [Widget Editor] Open new widget after creation
when creating a new widget, open it right away. 

right now, you manually need to open the newly created widget.

Corepex:
Resolves #2973 

## Additional info
This pull request updates the widget creation flow to ensure that the newly created widget's unique `id` is used throughout the process, rather than its `name`. This improves consistency and allows for more reliable widget handling after creation. The most important changes are:

**Widget Creation Callback Interface:**

* Changed the signature of the `createWidget` function and its callback so that it now passes the widget's `id` (a string) instead of its `name` to the `onFinish` callback. This affects both the type definition and the implementation. [[1]](diffhunk://#diff-8ab23b20c2379915a3a8e259e028be7f5eee5e23ce58988b267c1af8d8c6fcb0L21-R21) [[2]](diffhunk://#diff-8ab23b20c2379915a3a8e259e028be7f5eee5e23ce58988b267c1af8d8c6fcb0L66-R66)

**Widget Creation Implementation:**

* Updated the `createWidget` function to extract the `id` from the API response after widget creation and pass it to the `onFinish` callback, instead of the widget's name.

**Widget Editor Provider Logic:**

* Updated `WidgetEditorProvider` to use the new `id` parameter in the callback, and to call `openWidget(id, widgetType)` after a widget is created, ensuring the correct widget is opened.

These changes make the widget creation flow more robust by consistently using the widget's unique identifier.


xIrusux (approved):
LGTM - pls accept sonar qube issues

PR #3333, opened 2026-04-13
Fixed `sortingFilter` param in gdpr extractor

### LINKED ISSUE #2885: [GDPR] sorting only works on the first page
The sorting of the columns only sorts the results of the first page.
When you navigate to the next page the sorting is incorrect.
The sorting should sort all results.

Corepex:
Resolves #2885

## Additional info
This pull request refines the sorting and pagination logic in the `GDPRDataExtractorContainer` component to improve user experience and state management. The main focus is on ensuring that changing the sort order resets the page to 1 and that the sort state is properly tracked and updated.

**Improvements to sorting and pagination logic:**

* Added a new `sortFilter` state using `useState` to track the current sort order.
* Updated the `executeSearch` function to:
  * Set the `sortFilter` state when a new sort is applied.
  * Reset the page to 1 when the sort filter changes and no explicit page override is provided, ensuring users always start at the first page after sorting.
  * Keep the current page unless a sort change occurs, in which case the page is reset.


xIrusux (approved):
The page reset on sort change makes sense and should fix the “sorting only works on the first page” behavior. I’d just suggest deriving nextSortFilter / nextPage first and using those for the request, then updating state afterward. That makes the flow easier to read and avoids mixing request and state updates.

I.e.  


```
const nextSortFilter = overrides?.sortFilter ?? sortFilter
const isSortChange = overrides?.sortFilter !== undefined && overrides?.page === undefined
const nextPage = isSortChange ? 1 : (overrides?.page ?? page)
const nextProvider = overrides?.provider ?? provider
const nextColumnFilters = overrides?.columnFilters ?? columnFilters
const nextPageSize = overrides?.pageSize ?? pageSize

if (overrides?.sortFilter !== undefined) {
  setSortFilter(overrides.sortFilter)
}
if (isSortChange) {
  setPage(1)
}

if (nextProvider === '' || isEmpty(nextColumnFilters)) return

void trigger({
```

PR #3378, opened 2026-04-22
Applied translations for " Permissions Bundles"

### LINKED ISSUE #2960: [Permissions] label for permissions is missing if permissions are set
Once you set permissions, the label for the 2 different categories (Permissions, Permissions bundle) is not shown anymore.
Is it possible to also show the label if permissions are set to make it easier to distinguish?

Image

vmalyk:
Applied translate for permissions label for the both sections and moved mapping function to variable. As I see translations have been present and need just small wrapping select options.   

Looks like it resolves #2960.


xIrusux (approved):
That's exactly what is missing. Thanks. 
This looks good to me.

PR #3394, opened 2026-04-22
[User Management] - allow saving dO and document folders in role workspaces without further configuration 

### LINKED ISSUE #3335: [User Management][Roles] Adding Workspace for DataObjects without setting all Properties throws Error
When adding a Path with a DataObject Folder for DataObjects in Workspaces for Roles, without setting all Checkboxes, leads to an Error `This value should be of type bool. This value should be of type bool.`

Image

When checking all Checkboxes, the saving works again. After that I can remove checks again and saving works still. 
Looks like the default value might be null instead of false?

Image

Image

- pimcore/pimcore:v12.3.4
- pimcore/studio-backend-bundle:v2025.4.0
- pimcore/studio-ui-bundle:v2025.4.0

xIrusux:
add missing save and unpulish defaults for document and dataObject workspaces in roles

Resolves #3335


Corepex (approved):
LGTM - resolver checks before merging ;)

PR #3400, opened 2026-04-23
Update disclosure widget UX

rjjackson22:
Updated the tree expander disclosure widget to be `chevron-down` when expanded and `chevron-right` when closed.

## Additional info
Several different design standards suggest using right/down chevrons for disclosure widget expanders, including [ARIA-APG](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/), [Ant Design](https://ant.design/components/tree), and [Apple](https://developer.apple.com/design/human-interface-guidelines/disclosure-controls#Disclosure-triangles).

### DIFF
--- a/assets/js/src/core/components/element-tree/expander/tree-expander.tsx
+++ b/assets/js/src/core/components/element-tree/expander/tree-expander.tsx
@@ -66,13 +66,13 @@ export const TreeExpander = ({ node, state }: TreeExpanderProps): React.JSX.Elem
                 ? (
                   <Icon
                     options={ { width: 16, height: 16 } }
-                    value="chevron-up"
+                    value="chevron-down"
                   />
                   )
                 : (
                   <Icon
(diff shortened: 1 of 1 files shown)


idaiv:
That has to be updated. The arrows close/expand were based on the accordion back then but down/right is what we have to update it to.

The icons can be found here: https://www.figma.com/design/Q71sBiSRmHUDZhIqwEMVkm/Icon-Library?node-id=723-2215&t=amhLlMWhFBcIGybj-4

markus-moser:
Thanks for this, @rjjackson22! 🙏

We're targeting `2025.4` for this change, so I've opened #3764 which builds on your commit and additionally applies the same down/right treatment to our second tree implementation (`tree-element` — used by the perspective editor, tag filters, select options, thumbnails, tags, user/role management and reports), so the behavior is consistent across all trees.

Direction confirmed by @idaiv. Closing this in favor of #3764.

PR #3464, opened 2026-04-30
[Grid] update tag selection in grid after deleting an item

### LINKED ISSUE #1521: [Grid] element selection does not get updated after deletion
**Steps to reproduce**

- Login as admin
- Open an asset folder
- Switch to the list view
- Select 4 objects
- Delete one of those objects

**Actual behaviour**
- Selector still shows 4 objects

Image

**Expected behaviour**

- Selector should be updated to show 3 objects

**Environment**

- Playground
- Windows
- Google Chrome

xIrusux:
Resolves #1521


Corepex (approved):
Please fix the sonar issues - besides that it looks good 💯

PR #3489, opened 2026-05-06
Updated email tab logic

### LINKED ISSUE #2954: [Sent Mails] Server Error when opening HTML preview of plain text mail
- send a test mail with just plain text, no HTML part
- open mail in `sent mails` and click on HTML preview --> triggers an http error

Image

either we solve that on server-side so that a proper response is returned
or we hide the html tab as a whole.

Corepex:
Resolves #2954

## Additional info
This pull request improves the handling and display of email logs in the email log UI, specifically ensuring that the text and HTML previews are only enabled when the corresponding logs are available. This prevents unnecessary API calls and disables UI elements when data is missing.

**UI behavior improvements:**

* The "HTML" tab in the email log UI is now disabled if the email does not have an HTML log available, preventing users from accessing a non-existent preview.
* The "Text" tab's preview component now receives a `hasTextLog` prop, and will not attempt to fetch or display a text log if it is unavailable. [[1]](diffhunk://#diff-92c8bfaba8979d1718486ec363a87b1c8faedd2b0d49345ad7db2e3f96b42b5cL40-R45) [[2]](diffhunk://#diff-918b9ba0e022dd7e607afa01909f91769d924985a00a95df48716bdc6f742ff8R21-R31)

**Performance and correctness:**

* The `TextPreview` component skips the API call to fetch the text log if `hasTextLog` is false, reducing unnecessary network requests.
* The empty state for the text preview is now shown if there is no text log, in addition to the previous check for empty data, providing more accurate feedback to the user.


xIrusux (approved):
LGTM :)

PR #3529, opened 2026-05-08
Made API url configurable

### LINKED ISSUE #729: [Config] Make API prefix configurable
* We need to make the configurable API prefix available in this helper:
https://github.com/pimcore/studio-ui-bundle/blob/209d3cedecfa9013191b8aa9016cb445b2498cae/assets/js/src/core/app/api/pimcore/route.ts#L18

The `baseUrl` is already configurable here but we need to respect it in our helper:
https://github.com/pimcore/studio-ui-bundle/blob/209d3cedecfa9013191b8aa9016cb445b2498cae/templates/default/index.html.twig#L22

* Additionally it needs to be checked if we use this prefix everywhere. It does not make sense to Präfix the API calls with the domain (relative paths are enough). So we need to check where the getDomain() helper is used:
https://github.com/pimcore/studio-ui-bundle/blob/209d3cedecfa9013191b8aa9016cb445b2498cae/assets/js/src/core/app/api/pimcore/route.ts#L14

         * Maybe we can even get rid of `getDomain()`? 
         * Search the whole project for "localhost" and replace hardcoded locations with our helper
         * Search the whole project for "/api" and "/pimcore-studio/api" and replace hardcoded locations with our helper

* A special problem are our generated API slices - we would need to apply our helper there somehow without creating a custom frontend build:
https://github.com/pimcore/studio-ui-bundle/blob/209d3cedecfa9013191b8aa9016cb445b2498cae/assets/js/src/core/modules/element/element-api-slice.gen.ts#L18

Corepex:
Resolves #729 

## Additional info
This pull request introduces dynamic API prefix support to the frontend and backend, allowing the API base path to be configured at runtime. This change improves flexibility for deployments where the API path may differ from the default. The most important changes are grouped below:

**Frontend API configuration:**

* Updated `getPrefix` in `route.ts` to return the API prefix from `appConfig` instead of a hardcoded string, enabling dynamic configuration.
* Modified the API base query in `index.ts` to use a dynamic base URL by replacing the default prefix with the runtime value from `getPrefix`.
* Added an optional `apiPrefix` property to the `AppConfig` interface to support the new configuration.

**Backend API prefix injection:**

* Injected the API prefix (`pimcore_studio_backend.url_prefix`) into the controller via Symfony's service configuration (`services.yaml`).
* Updated the controller to include the `studioApiPrefix` in the frontend app configuration sent to the client, ensuring the frontend receives the correct API prefix.


assets/js/src/core/app/api/pimcore/index.ts
```diff
@@ -8,12 +8,23 @@
...
+import { getPrefix } from './route'
 
-const baseQuery = fetchBaseQuery({ baseUrl: '/' })
-export type BaseQuery = typeof baseQuery
+const dynamicBaseQuery: BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError> = async (args, api, extraOptions) => {
+  const prefix = getPrefix()
+  const defaultBaseUrl = '/pimcore-studio/api'
+
+  const adjustedArgs = typeof args === 'string'
+    ? args.replace(defaultBaseUrl, prefix)
```
markus-moser: just to be super-safe, should we do something which guarantees that it's only replaced at start of the base url?
```typescript
const replaceBaseUrl = (url: string): string =>
      url.startsWith(defaultBaseUrl)
        ? `${prefix}${url.slice(defaultBaseUrl.length)}`
        : url
```

markus-moser (changes requested):
LGTM
just to be really safe, could you please adjust my single comment?

PR #3546, opened 2026-05-11
Added translation debug processor

### LINKED ISSUE #3126: [System settings] Debug Admin-Translations (wrapped in +) has no function
Image

in classic UI the labels were wrapped in + + 

Image

either implement that function or remove the setting.

Corepex:
Resolves #3126 

## Additional info
This pull request enhances the i18n (internationalization) infrastructure by introducing a debug mode for translations. This allows developers to visually identify when translation debugging is enabled, making it easier to spot translation issues during development. The most important changes are grouped below:

**Debug Translations Feature:**

* Added a new `createDebugTranslationsProcessor` post-processor in `post-processors.ts` that wraps translations with `+ ... +` when debug mode is active.
* Integrated the debug translations processor into the i18n setup in `index.ts`, using the Redux store to determine if debug mode is enabled via the `debug_admin_translations` setting. [[1]](diffhunk://#diff-0b4024656b2848c25a63406f4f6a01f515bd196f315646228892954153fecbc9L15-R16) [[2]](diffhunk://#diff-0b4024656b2848c25a63406f4f6a01f515bd196f315646228892954153fecbc9R39-R43)
* Updated i18n initialization to include the new `debugTranslations` post-processor alongside the existing `returnKeyIfEmpty` processor.


assets/js/src/core/app/i18n/utils/post-processors.ts
```diff
@@ -12,6 +12,18 @@
...
  */
+export const createDebugTranslationsProcessor = (getDebugMode: () => boolean): {
+  type: 'postProcessor'
+  name: string
+  process: (value: string) => string
+} => ({
+  type: 'postProcessor' as const,
+  name: 'debugTranslations',
+  process (value: string): string {
+    return getDebugMode() ? `+ ${value} +` : value
```
markus-moser: Just wondering - do we show "+ Translation Key +" in that case or "+ Translated Text +"? And are we sure that we don't pollute our translation table with additional "+ ... +" translatin keys?

markus-moser (commented):
In general it LGTM, maybe you can quickly check my comment and explain how it works. Could also have a short call about it if helpful.

assets/js/src/core/app/i18n/utils/post-processors.ts
Corepex: It shows `+ Translated Text +`. The `process(value)` function in a post-processor receives the already-resolved translation value — i18next runs the full key lookup first, then passes the result through post-processors. So the + wrapping is applied to the final translated string.

markus-moser (approved):

PR #3550, opened 2026-05-11
Translated selectable options in relation-editable

### LINKED ISSUE #3355: AdvancedManyToManyObjectRelations the options are not translatable
If you use many to many object relations and define type select or multiselect

<img width="1350" height="481" alt="Image" src="https://github.com/user-attachments/assets/023a8b38-fad0-48d3-bdcd-77de1c668670" />

The options are currently not translatable, see:

https://github.com/pimcore/studio-ui-bundle/blob/2026.x/assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/helpers/relations/hooks/use-convert-relation-editable-columns.ts#L49

https://github.com/pimcore/studio-ui-bundle/blob/2026.x/assets/js/src/core/modules/element/dynamic-types/definitions/objects/data-related/helpers/relations/hooks/use-convert-relation-editable-columns.ts#L59

<img width="1806" height="505" alt="Image" src="https://github.com/user-attachments/assets/a79642c6-3264-4bd5-8c29-5dff856c0f22" />

Needs to do something like this:

// For select:                                                                                                                                                                                                                   
  config: {                                                 
    options: value?.split(';').map(v => ({ value: v, label: t(v) })) ?? []
  }   

Classic Ui Options are translatable.

Corepex:
Resolves #3355

## Additional info
This pull request updates how editable columns are generated for relation data types, specifically improving the way select and multi-select column options are handled to support translations. The main change is that option labels for select and multi-select columns are now passed through the translation function, ensuring localized display for users.

**Improvements to column option localization:**

* The `mapColumnType` function now accepts a translation function `t` as an argument and uses it to translate option labels for both select and multi-select column types. [[1]](diffhunk://#diff-14940dfd58c2094bc171ef3a03110678059322add73028698666ba2bab5ce7c0L29-R29) [[2]](diffhunk://#diff-14940dfd58c2094bc171ef3a03110678059322add73028698666ba2bab5ce7c0L49-R49) [[3]](diffhunk://#diff-14940dfd58c2094bc171ef3a03110678059322add73028698666ba2bab5ce7c0L59-R59)
* The call to `mapColumnType` within the `useConvertRelationEditableColumns` hook has been updated to provide the translation function `t` as an argument.


xIrusux (approved):
LGTM :)

PR #3554, opened 2026-05-11
Close grid config modal on cancel

### LINKED ISSUE #3436: [Data Object][Grid][Advanced Columns] Grid config modal cancel button has unexpected function
right now, it resets the config changes, but does not close the modal

Image

Corepex:
Resolves #3436 

## Additional info
This pull request makes a small update to the behavior of the Cancel button in the `GridConfigModal` component. Now, when the Cancel button is clicked, it not only triggers the `onCancelClick` handler but also ensures the modal is closed by calling `onOpenChange` with `false` if that callback is provided. 

([assets/js/src/core/modules/data-object/listing/decorator/column-configuration/view-layer/components/grid/hooks/use-grid-options/tabs/grid-config/grid-config-modal.tsxL101-R101](diffhunk://#diff-142a80b0f25fe6b62127cfcf6b4c559a93853230e6b4a24d45b39222a1eff5b5L101-R101))


xIrusux (approved):
Lgtm :)

PR #3562, opened 2026-05-12
[Cache] - add Menu entry and options to clear cache in different ways

### LINKED ISSUE #3491: [Megamenu] Cache Menu Entries are missing
Image

xIrusux:
Resolves #3291


assets/js/src/core/modules/cache/cache-actions.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import { store } from '@Pimcore/app/store'
+import { ApiError, trackError } from '@sdk/modules/app'
+import { api } from './cache-api-slice.gen'
+import i18n from 'i18next'
+import { message } from 'antd'
+import { isUndefined } from 'lodash'
+
+const showSuccess = (key: string): void => {
```
Corepex: i get the idea of this function but i would (personally) use `void message.success(i18n.t(key))` directly

Corepex (approved):
LGTM - good job

assets/js/src/core/modules/cache/cache-actions.ts
xIrusux: done

PR #3572, opened 2026-05-13
[Bug] Bricks should be saving class name and not id.

lukemacausland:
This changes the logic to save the class using name instead of class id, so its the same behavior as the legacy admin. This also fixes multiple other small errors / bugs when working within the brick editor in studio.

You can reproduce this in the demo by:
1. Opening up a brick
2. Attempt to save the brick to a class with a mismatch class id and name
3. Hit save

image
image

### DIFF
--- a/assets/js/src/core/modules/object-brick/components/object-brick-editor/object-brick-class-definitions-block.tsx
+++ b/assets/js/src/core/modules/object-brick/components/object-brick-editor/object-brick-class-definitions-block.tsx
@@ -22,6 +22,7 @@ import { useTranslation } from 'react-i18next'
 
 interface FieldnameSelectProps {
   blockIndex: number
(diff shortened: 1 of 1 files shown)


lukemacausland:
@vin0401 @markus-moser - Just wondering if this is a potential pull request that will be reviewed / merged?

markus-moser:
@lukemacausland Thanks a lot for your PR! We will check the PR in detail very soon... We need to analyze it in detail but if there is a bug we should definitely fix it!

lukemacausland:
Thanks @markus-moser for the update

vin0401:
Hi @lukemacausland,

we checked it now and accepted your changes in this PR: https://github.com/pimcore/studio-ui-bundle/pull/3619
Thanks a lot!

PR #3611, opened 2026-05-22
Fix for significantly improved loader/background performance

rjjackson22:
When studio ui is open in a browser it consumed up to 100% GPU utilization, make the app slow and unresponsive. It can also cause other apps on the client machine that require the GPU for rendering to become sluggish. Turning off browser hardware acceleration ruins performance, dropping the app to single digit FPS. This behaviour exists in both Chrome and Firefox, but is much worse in Firefox.

Steps to recreate:
1. Disable hardware acceleration in Chrome/Firefox
2. Navigate to https://demo.pimcore.com/pimcore-studio (or any other instance running 2026.1)

Technical problem(s):
- The loading screen saturated the GPU at 100% during normal usage. Five simultaneously animated elements used backdrop filter: blur(20px) on a full-viewport overlay and filter: blur(90–310px) on large elements, combined with a background keyframe animation that forces CPU repaints on every frame. 
- The handoff from the Twig preloader to the React component caused a visible judder, and the fade transition itself was choppy because it started while the main thread was busiest.

Changes:
  - Removed backdrop-filter: blur(20px) from both the preloader template and React styles; replaced with a cheap rgba overlay.
  - Reduced blur radii on background figures.
  - Removed pimcore-color-pulse background-color keyframe animation.
  - Simplified React background.styles.ts to always render the idle state.
  - Gated the preloader fade on phase === 'idle' in background.tsx so the transition starts only after the main thread has finished its initial render.

Video example of bad performance:

https://github.com/user-attachments/assets/e8dd7bbd-7f51-4f8c-ba10-253790f2b8ec


vin0401:
Hi @rjjackson22,

thanks for your PR and also for bringing that topic up. 
Never the less we decided for a bit different approach: https://github.com/pimcore/studio-ui-bundle/pull/3616

- Keeping the color pulse intact
- Move the blurring of the colored figures completely to the backdrop filter
- Scaling of the overall animation size to the average viewport
- Make sure that idle really means idle
- Added support for the prefer-reduced-motion flag

Overall that brought the GPU usage from ~ 95% (left peak) down to ~ 27% (right peak).

image

This will also be part of the todays bug fix release.
Thanks again for bringing it up!

PR #3628, opened 2026-05-27
Further performance enhancements for background loader

rjjackson22:
While the fixes from https://github.com/pimcore/studio-ui-bundle/pull/3616/ do improve performance and reduce peak/idle GPU utilization, we're still noticing sluggish UI - mostly choppy tab animations and laggy on hover highlighting in the element trees. When hardware acceleration is turned off (to emulate client devices with no dedicated GPU), the app is still unusable.

Suspected Cause:
- The `backdrop-filter: blur(90px);` in the `background.styles.ts` seems to be the main culprit. When it's removed, the UI runs buttery smooth and the GPU idles at ~0%.

Solution:
- Remove the `backdropBlur` element in favour of an increased `filter: blur(90px)` on the background-figure elements themselves. The animation and background look virtually identical.
- Additionally, the `will-change: transform` only really needs to be on the background figures while they are orbiting - once the app is loaded they won't change anymore and don't need to be tracked. This was actually causing some extra performance issues with no GPU. [MDN actually recommends against](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/will-change) using `will-change` if possible, so I think restricting it to only during the orbit animation is good enough.

### DIFF
--- a/assets/js/src/core/components/background/background.styles.ts
+++ b/assets/js/src/core/components/background/background.styles.ts
@@ -92,8 +92,8 @@ export const useStyle = createStyles(({ css }, { phase, backgroundShade }: Style
 
       .background-figure {
         position: absolute;
-        will-change: transform;
         flex-shrink: 0;
+        filter: blur(90px);
 
         ${isOrbiting
           ? css`
@@ -198,13 +198,6 @@ export const useStyle = createStyles(({ css }, { phase, backgroundShade }: Style
         }
       }
     `,
-    backdropBlur: css`
-      position: absolute;
-      inset: 0;
-      backdrop-filter: blur(90px);
-      -webkit-backdrop-filter: blur(90px);
-      pointer-events: none;
-    `,
     logoImage: css`      position: absolute;
       top: 50%;
       left: 50%;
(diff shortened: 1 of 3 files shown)


rjjackson22:
@vin0401 thank you for looking at my last PR, your fixes look like what we need for the most part. We found some additional performance enhancements though, could someone take a look at this whenever there's a free moment?

vin0401:
Hi @rjjackson22,

I had another look at it. And I agree the backdrop filter has some downsides here.
So we decided for a more combined approach: https://github.com/pimcore/studio-ui-bundle/pull/3631

That means, we stick with the backdrop filter only during the animation phase.
It just creates the visual smoother result and also seems to be less expensive then blurring multiple items separately.

For the actual Idle phase we then stick to your approach, so static blurred items to prevent the constant recomputing of the background.

During the tests with hardware acceleration disabled animation seemed to be smooth now. 
And looking at some metrics during normal usage show quite some improvement.
In this case it's the hovering over the tree, and the cost to trigger the detail tooltip:

image

Thanks again for the efforts!

PR #3629, opened 2026-05-27
Bug fix: infinite render loop for objects with manyToManyObjectRelation fields

rjjackson22:
Whenever a data object with a `manyToManyObjectRelation` field is opened in studio ui the UI gets sluggish and the browser starts continuously firing the following console error:
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
ManyToManyObjectRelationInner@http://localhost:3030/static/js/async/__federation_expose_app.js:111615:68
DynamicTypeRegistryProvider@http://localhost:3030/static/js/async/__federation_expose_app.js:125673:36
ManyToManyObjectRelation
```

Solution:
- There seems to be an infinite loop, memoizing data in `use-data-object-grids.ts` seems to fix it.


ValeriaMaltseva:
@rjjackson22
Hi! Thank you for this contribution - it solves the issue we were hitting. Really appreciate your help! 🙌

ValeriaMaltseva (approved):

PR #3657, opened 2026-06-03
fix: role editor multiselect renders selected values

Pippo433:
The role settings form seeded the `perspectives` multiselect with the API's
PerspectiveConfig objects, but the Select options use the perspective id as their
`value`. antd can't match objects to option values, so saved perspectives showed
as empty chips after reload. Map the objects to their ids, like `classes` already does.


Pippo433:

xIrusux:
@Pippo433 Thanks for the PR and spotting this! It has since been fixed and is available to you in our latest release.
Take a look at [https://github.com/pimcore/studio-ui-bundle/pull/3688](url) for the complete fix.

PR #3683, opened 2026-06-09
[WidgetManager] Extract WidgetManagerService as injectable DI service

### LINKED ISSUE #2577: Add Widget Manager as a service to the SDK
Currently, to use the widget manager methods in bundles inside the class, you can only access them through the store. It would be great to add the ability to use these methods directly by injecting the widget manager as a service.

ValeriaMaltseva:
Resolves #2577


assets/js/src/core/app/config/services/service-ids.ts
```diff
@@ -78,6 +78,7 @@ export const serviceIds = {
 
   // Widget manager
   widgetManager: 'WidgetManagerService',
+  widgetManagerService: 'WidgetManager/ActionService',
```
markus-moser: This constellation is a bit strange. The widgetManager is called "WidgetManagerService" and the widgetManagerService "WidgetManager/ActionService". Not sure how to solve it the best way, but if we have no other idea maybe renaming the `WidgetManagerService` to `WidgetMananagerActionService`?

markus-moser (changes requested):
In general it LGTM but we should maybe double check the naming conventions :)

PR #3685, opened 2026-06-09
[Document Editor][Image Editable] Respect minWidth/minHeight config option

### LINKED ISSUE #2343: [Document Editor][Image Editable] Respect minWidth/minHeight config option

ValeriaMaltseva:
Resolves #2343


markus-moser (changes requested):
This doesn't fully resolve #2343 yet — `minWidth`/`minHeight` should work like in the classic UI, where they are a validation constraint, not just styling: after assigning an image, the dimensions are checked and if the image is too small, the value is rejected (editable emptied) and an "image is too small" error is shown ([classic image.js, ~L432–450](https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/public/js/pimcore/document/editables/image.js#L432-L450)).

Missing:
- Validate dimensions against `minWidth`/`minHeight` on assign (drop / search / upload)
- If too small: reject the value and show an error (classic: `image_is_too_small`)

markus-moser (approved):

PR #3723, opened 2026-06-11
UI polish of the user menu and avatar instances

fashxp:
Resolves #


assets/js/src/core/modules/app/base-layout/user-menu/user-menu.tsx
```diff
@@ -65,12 +67,15 @@ export const UserMenu = ({ className }: IUserMenuProps): React.JSX.Element => {
...
 
+  const notificationCount = data?.unreadNotificationsCount ?? 0
+
   const items: DropdownMenuProps['items'] = [
     {
       key: 'title',
       label: (
         <div className={ 'user-menu__title' }>
-          {t('user-menu.title')}
+          PROFILE
```
vin0401: We would need to keep the old translations key here.

assets/js/src/core/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import React from 'react'
+import { Avatar as AntAvatar, type AvatarProps as AntAvatarProps } from 'antd'
+import { Icon } from '@Pimcore/components/icon/icon'
+
+export interface AvatarProps extends AntAvatarProps {}
+
+const getDefaultIconSize = (size: AvatarProps['size']): number => {
+  if (typeof size === 'number') return Math.round(size / 2)
```
vin0401: The calculation for the size looks a bit strange here. Is it it itentional that it get's halved.

vin0401 (changes requested):
Overall it looks pretty good already. Nice first PR!!!
Still we would to adjust two things.

vin0401 (approved):

PR #3749, opened 2026-06-12
[Appearance & Branding][Grid] Disable read-only form fields and update selection after batch edit

### LINKED ISSUE #2937: [Appearance & Branding] configuration in read only mode can be modified
**Steps to reproduce**
- Make sure the configuration for 'Apperance & Branding' is read only
- Login as admin
- Change some config options (e.g. some color)

**Actual behaviour**
- Options can be changed allthough config is read only
- Configuration can not be saved

**Expected behaviour**
- If the configuration is read only. it should not be possible to change any of the options in the configuration

**Environment**

- Studio Nightly
- Windows

### LINKED ISSUE #3477: [Grid] After batch edit on filtered list number of selected objects is not updating
Steps to reproduce:

- filter out grid by some attribute
- select couple of objects in grid
- batch edit filtered attribute
- observer number of selected objects

Expected: selected number should update since it is confusing for user

https://github.com/user-attachments/assets/48bc9cac-164d-4146-8537-1e2983d1eaf7

xIrusux:
### What

Two fixes:

**[Appearance & Branding] configuration in read-only mode can be modified** (fixes #2937)
- The form already disabled the Save button via `adminSettings.writeable`, but the fields themselves stayed editable.
- `ColorPanel` and `ImagePanel` now accept a `disabled` prop, and `AppearanceForm` passes `!isWriteable` down, disabling both color pickers and both image pickers when the config is not writeable.

**[Grid] selected-object count not updating after batch edit on filtered list** (fixes #3477)
- The batch edit jobs never refreshed the grid or reconciled the selection on completion — `refreshGrid` was only wired to the notification "reload" button, so rows that fell out of the active filter stayed counted in "x selected".
- The data object and asset batch edit modals now pass an `onFinish` callback (same pattern as batch delete in `batch-actions.tsx`): refresh the grid and, for the selected-rows path, clear the selection.

### Testing

- `npm run lint` (changed files) and `npm run check-types` pass.
- Manual: see issue repro steps — after batch-editing filtered rows, the grid reloads and the selection count resets; with a read-only admin config, all Appearance & Branding fields are disabled.


assets/js/src/core/modules/data-object/listing/batch-actions/batch-edit-modal/batch-edit-modal.tsx
```diff
@@ -194,7 +197,11 @@ export const BatchEditModal = ({ batchEditModalOpen, setBatchEditModalOpen }: Ba
         selectedRowsIds: selectedRowsIds.map(Number),
         values,
         assetContextId: id,
-        refreshGrid
+        refreshGrid,
+        onFinish: async () => {
+          await refreshGrid()
+          setSelectedRows({})
```
markus-moser: just clearing the row selection is not a good idea. let's revert those changes, we would need to think about this issue in detail if we plan to solve it ;)

markus-moser (changes requested):

PR #3758, opened 2026-06-12
[Document Editor] Add areablock copy/paste feature

### LINKED ISSUE #1998: [Document Editor] Add areablock copy/paste feature

ValeriaMaltseva:
Resolves #1998


markus-moser (changes requested):
LGTM and works well. Really like that it's now also possible to copy/paste across different browser windows due to local storage.

The only thing that's missing is the cut (/paste) feature. Can you also add it?

I'm also wondering if we should introduce tooltips for the buttons. That could maybe be a follow-up, as it would make sense to add tooltips to the other buttons as well. I'd say it would be a real benefit, as otherwise it's a bit of guessing (is this the paste button? Is it the copy button?).

Issue #2886, opened 2026-02-04
[GDPR] export should not work for elements without permission


heigpa:
**Steps to reproduce**
- Login as admin
- Create or adapt a user and exclude some data objects
- Login as this user
- Search for those objects in the GDPR extractor
- Export one of the data objects

**Actual behaviour**
- Data object can be exported without permission for this data object

**Expected behaviour**
- It should not be possible to extract data of a data object without permission

**Environment**

- Studio Nightly
- Windows
- Google Chrome

Corepex:
Since I rely on the `__gdprIsDeletable` flag the backend provides us, I would ask @martineiber to have a closer look at that.

heigpa:
It works now but the error message could be improved. I changed the label to improvement and will reopen the issue.

<img width="1178" height="943" alt="Image" src="https://github.com/user-attachments/assets/a4403a4f-bcc4-4b67-9bd9-49a7ac3a0547" />

lukmzig:
Should be resolved now

heigpa:
looks good. thx

PR #3773, opened 2026-06-18
[GDPR]: export error message

lukmzig:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/2886

## Additional info
- add missing error message string
- fix - error modal opens twice for each error


markus-moser (approved):
Thanks a lot, LGTM.

Just one note:
In general we decided to use `trackError()` in most situations only for API errors as each call results in a error modal for the user and we don't wan't to bring up each small runtime error as error modal as otherwise the user might be flooded with errors.

So normally we do it like described in the skill in most situations:
https://github.com/pimcore/skills/blob/main/skills/pimcore-studio-ui-error-handling/SKILL.md#the-standard-pattern-rtk-query--useeffect--trackerror

I'd say for this case it's also fine like you did it, but just as a hint for the future and also that you are aware of the error handling skill.

assets/js/src/core/modules/gdpr-data-extractor/components/export-button/export-button.tsx
```diff
@@ -37,15 +36,12 @@ export const ExportButton = ({ id, providerKey, onClick, loading, ...iconButtonP
       onClick?.(e)
     } catch (error: any) {
       console.error('Export failed:', error)
-      trackError(new GeneralError(error.message as string))
```
markus-moser, line 40: In other words, would have been also ok to just remove this line.

lukmzig:
> Thanks a lot, LGTM.
> 
> Just one note: In general we decided to use `trackError()` in most situations only for API errors as each call results in a error modal for the user and we don't wan't to bring up each small runtime error as error modal as otherwise the user might be flooded with errors.
> 
> So normally we do it like described in the skill in most situations: https://github.com/pimcore/skills/blob/main/skills/pimcore-studio-ui-error-handling/SKILL.md#the-standard-pattern-rtk-query--useeffect--trackerror
> 
> I'd say for this case it's also fine like you did it, but just as a hint for the future and also that you are aware of the error handling skill.

Thanks for the info! I actually did not use skills here which is another good point, would have probably lead to a solution which aligns more with the standard approach. I will keep that in mind for the future.

Issue #3373, opened 2026-04-21
Headline vs Toolbar - component rework


idaiv:
There are inconsistencies in the usage and styling of the toolbar component. This issue was faced when reviewing the implementation of the Email list - @Corepex has informed me that the headline uses the toolbar component. In my opinion these should be individual components to fulfill the requirements of each case. 

Storybook component: https://studionightly.pimcore.app/storybook/?path=/docs/components-layout-toolbar--docs

Areas of focus:
- Top bar - breadcrumbs and tabs (open element)
- Top bar - breadcrumbs and draft version (detached tab/ bottom widget)
- Headline - in all the config screens and additional tabs (Emails, Reports, Dependencies, Metadata, etc.) - some have simply a headline, others have additional buttons, selects, segmented tabs, etc.
- Toolbar - where we have the pagination and additional left buttons, rows selection, etc. 

The problem is that all of these areas are individual design components on my side and they follow certain heights specifically fitting based on their type. However on the implementation side, these areas heights do not match with the design. 

Listed below are the differences:
- Top bar - open element: Design 40 px; Implementation 48 px
Improvement suggestion: Reduce padding top and bottom of the breadcrumb element from 8 px to 4 px

- Top bar - bottom widget: Design 40 px; Implementation 40 px
Improvement: The content needs to be center aligned vertically. Currently it is with a fixed height and therefore the paddings of the breadcrumbs are not respected and the visualisation is messed up.

- Headline - different cases: Design 44 px; Implementation differs between 48 px to 52 px (what I have found so far)
Since there are quite some variations of what can be in the headline, there will be a provided design link with an overview. 

- Toolbar - pagination and selection: Design 48 px; Implementation 47 px? (the pagination seems to be 31 px and has padding top + bottom 8px)

For more details on the implementation and directions, please reach out to @vin0401 as we have done the review together.

idaiv:
discussed with @vin0401. We will not split up the component into toolbar and header. We will rather try to clean up the paddings, height etc for the already existing variants to provide a more consistent implementation. 

To do front end team: 

- [ ] Review your variants of the Toolbar component - do we need left and right aligned? Why is there one for size specifically. Please, clear this out so each variant is one we use and makes sense to be there.
[...]

PR #3774, opened 2026-06-18
Updates on the toolbar component implementation

idaiv:
Resolves #https://github.com/pimcore/studio-ui-bundle/issues/3373

## Additional info
- toolbar.styles.tsx — Base flex layout, headline size class, none position class
- toolbar.tsx — Added 'headline'/'none' to types, smart defaults for headline
- toolbar.stories.tsx — Renamed stories, fixed demo data, added HeadlineWithBorder
- select.styles.ts — Multi-select padding 2px → 1px (32px height fix)
- editor-tabs.tsx — Breadcrumb padding extra-small → mini (48px → 40px)
- translation-merger.tsx — Buttons relocated, Pagination swapped, top bar simplified
report-top-bar.tsx — Unwrapped from Content, size='auto' → size='headline'
- appearance-form.tsx — Headline toolbar added, Title level={2} removed from content
- collections-tab.tsx (×2) — Raw Flex → Toolbar, Header → Title, gap + SearchInput props fixed
- 17 screens (no border) — Verbose props → <Toolbar size='headline'>
- 5 screens (with border) — Verbose props → <Toolbar position='top' size='headline'>


assets/js/src/core/components/toolbar/toolbar.tsx
```diff
@@ -20,8 +20,8 @@ export interface ToolbarProps {
   align?: FlexProps['align']
   theme?: 'primary' | 'secondary'
   borderStyle?: 'default' | 'primary'
-  position?: 'top' | 'bottom' | 'content'
-  size?: 'small' | 'auto' | 'default'
```
vin0401, line 24: I think it would make sense to move that one inside of a new component already.
Since headline seems really like something different compared to what we normally have in the toolbar sizes. 

And it would also be beneficial if we can control those in the future separately.

assets/js/src/core/components/toolbar/toolbar.stories.tsx
```diff
@@ -107,7 +109,7 @@ export const Size: Story = {
   }
 }
 
-export const Position: Story = {
+export const PositionTop: Story = {
```
vin0401, line 90: The Pipeline has thrown an error on that. Either we should call it just Position again, or we should the referencing story.

assets/js/src/core/components/toolbar/headline.tsx
```diff
@@ -0,0 +1,12 @@
...
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+// @deprecated — use Header from '@Pimcore/components/header/header' instead
+export { Header as Headline, type HeaderProps as HeadlineProps } from '@Pimcore/components/header/header'
```
vin0401: I think we can remove this.

PR #3776, opened 2026-06-18
[Documents] - highlight editable elements is missing

### LINKED ISSUE #3390: [Documents] - highlight editable elements is missing
Image

ValeriaMaltseva:
Resolves #3390


assets/js/src/sdk/modules/document/index.ts
```diff
@@ -45,7 +45,7 @@ export * from '@Pimcore/modules/document/editor/sidebar/visibility/document-perm
 
 export * from '@Pimcore/modules/document/editor/widget'
 
-export * from '@Pimcore/modules/element/dynamic-types/definitions/document/editable/components/inheritance-overlay/inheritance-overlay'
+export * from '@Pimcore/modules/element/dynamic-types/definitions/document/editable/components/editable-overlay/editable-overlay'
```
markus-moser, line 48: That's a BC break. We need to find a way that we still provide the InheritanceOverlay. Maybe as alias? (we could remove this alias in future version so maybe already add a deprecation comment)

markus-moser (changes requested):
In general it LGTM but we need to find a way to avoid the BC break.

PR #3840, opened 2026-07-02
Show version coauthor tag in versions timeline

fashxp:
## Summary

The Versions tab timeline shows a "Co-authored by <name>" tag (coauthor type in a tooltip) on versions carrying the new coauthor fields — shared component, so it applies to data objects, assets, and documents alike.

Part of the version-coauthor feature: core pimcore/pimcore#19235, API pimcore/studio-backend-bundle#1928, stamping pimcore/pimcore-agent-bundle#73. This PR is standalone-safe (optional fields, additive types); the tag simply never renders until the backend ships the fields.

## Changes

- OpenAPI snapshot: `Version` schema gains `coauthorType`/`coauthor` (hand-patched into the snapshot — a wholesale refresh currently fails codegen on an unrelated pre-existing `$ref` bug in the live spec: `BundleAgentCreateSampleParameters` is emitted without the `#/components/schemas/` prefix); regenerated `version-api-slice.gen.ts` (diff is exactly the two optional fields).
- `create-version-accordion-item.tsx`: `Extra` renders the coauthor `Tag` (SDK `Tooltip` wrapper showing the type) alongside the existing Published/Autosaved tags, preserving their precedence.
- New i18n key `version.coauthored-by` (en + de).

## Verification

`npm run lint-fix`, `npm run check-types`, `npm run build-app` all green. E2E on the enterprise demo: agent-created version shows the tag + "agent" tooltip; human versions show none.

https://claude.ai/code/session_01SsEa2MTm2gRPyGEdHGnnoF


translations/studio.de.yaml
```diff
@@ -439,6 +439,7 @@ version.label: Versionen
 version.versions: Versionen
 version.version: Version
 version.publish: Veröffentlichen
+version.coauthored-by: 'Mitverfasst von {{type}} {{name}}'
```
ValeriaMaltseva, line 444: Shouldn't it be added for the other locales as well? At the moment, it's only available for en and de :)

assets/js/src/core/modules/element/editor/shared-tab-manager/tabs/versions/helpers/create-version-accordion-item.tsx
```diff
@@ -93,11 +93,25 @@ export const createVersionAccordionItem = ({
 
   const Subtitle = (): React.JSX.Element => {
     const { t } = useTranslation()
+    const isCoauthored = !isNil(version.coauthor) && version.coauthor !== ''
```
ValeriaMaltseva: Here we could use the isEmptyValue helper from utils, as it already checks for isNil and !== ''. So this could simply be:

`const isCoauthored = !isEmptyValue(version.coauthor)`

ValeriaMaltseva (changes requested):
Generally, LGTM! 🚀
The only thing I'd like to double-check is the locales.

pimcore-deployments:

**Added (1):**
- `version.coauthored-by`


146 validation error(s) predating this PR were present at the merge base. This run also fixed 128 of them; 18 still remain. The remainder is not blocking — later runs keep chipping at it.

PR #3890, opened 2026-07-14
[Studio UI] Consistent save button order — Save Draft left, Save & Publish right

idaiv:
Solves: https://github.com/pimcore/service-operations/issues/924 and https://github.com/pimcore/platform-version/issues/198, fyi @astapc 

## Summary

Ensures consistent button positioning in the editor toolbar for both data objects and documents, regardless of published/unpublished state.

### Before
| State | Left button | Right button |
|-------|------------|-------------|
| Published | Save Draft (default) | Save & Publish (primary) |
| Unpublished | Save & Publish (default) | Save Draft (primary) |

Button positions **swapped** based on state — confusing for users who build muscle memory for button positions.

### After
| State | Left button | Right button |
|-------|------------|-------------|
| Published | Save Draft (default) | Save & Publish (primary) |
| Unpublished | Save Draft (default) | Save & Publish (primary) |

- **Save & Publish** is always the rightmost button with `type="primary"` styling
- **Save Draft** is always on the left with `type="default"` styling
- When unpublished, Save & Publish also triggers `publishDraft()` on success
- Folder type unchanged — shows only a single "Save" button
- Keyboard shortcuts unchanged
- Permissions unchanged

### Files changed
- `data-object/editor/toolbar/save-buttons/save-buttons.tsx`
- `document/editor/toolbar/save-buttons/save-buttons.tsx`


assets/js/src/core/modules/data-object/editor/toolbar/save-buttons/save-buttons.tsx
```diff
@@ -112,22 +112,18 @@ export const EditorToolbarSaveButtons = (): React.JSX.Element => {
         )
       }
 
-      const saveDisabled = isLoading || isSchedulesLoading || isDraftLoading
-
-      if (dataObject?.published === false && checkElementPermission(dataObject?.permissions, 'save')) {
+      if (dataObject?.published === false) {
```
vin0401: Seems like the permission check got lost here.

assets/js/src/core/modules/document/editor/toolbar/save-buttons/save-buttons.tsx
```diff
@@ -135,22 +135,18 @@ export const EditorToolbarSaveButtons = (): React.JSX.Element => {
         )
       }
 
-      const saveDisabled = isLoading || isSchedulesLoading || isDraftLoading
-
-      if (document?.published === false && checkElementPermission(document?.permissions, 'save')) {
+      if (document?.published === false) {
```
vin0401: Same here permission is missing.

vin0401 (changes requested):
We should check if the user has the save permission, otherwise we drop the button, I guess?

vin0401 (approved):

PR #3893, opened 2026-07-15
[Batch Edit] Persist every edited language for localized data object fields (#2492)

### LINKED ISSUE #2492: [Grid][Batch editing] edit several languages at the same time
Would it make sense to edit multiple languages in a batch edit mode in one go?
Actually you can only edit the language that is shown in the language selector of the field.
If you switch to another language, only the language that is actually shown will be saved.

Image

xIrusux:
Fixes #2492

## Problem
Batch-editing a **localized** data object field only persisted the language selected at save time. Editing e.g. German and English submitted only `{"localizedfields":{"name":{"de":"…"}}}`.

**Root cause:** the language selector swapped a single `Form.Item`'s `name` between languages. antd/rc-field-form keys the field by its name path, so a name change unmounts the old-language field; at submit, `getFieldsValue()` only collects **currently-registered** fields — so every language except the last-selected was dropped.

## Fix
- **data-object-adapter-component.tsx** — builds per-language entries via `useBatchEditLocales`
- **batch-edit-form-item.tsx** — renders one mounted `Form.Item` per language (only selected visible); no `initialValue` on localized items so untouched languages aren't submitted
- **use-batch-edit-locales.ts, get-batch-edit-field-name.ts** — the helpers it relies on
- **classification-store + object-brick** — updated to the new `items` signature (single entry, behavior unchanged)

Editing multiple languages now yields `{"localizedfields":{"name":{"de":"…","en":"…"}}}`; languages the user didn't touch are omitted (not overwritten with null).

## Scope
Reverted six unrelated generic batch-edit component changes to HEAD and dropped a now-unused wrapper, keeping the fix scoped to the data-object adapter path. `tsc` passes with no dangling references.

## Notes / risks
- Frontend build is regenerated by CI (`package-build` archive); no build artifacts committed.
- Advanced classification-store / object-brick localized fields keep their prior single-language behavior (out of scope for #2492).
- Non-data-object (asset metadata) generic components are back to flat `name={key}` — separate concern if per-locale is ever needed there.


xIrusux:
Superseded by #3895 (targets and based on `2026.x` instead of `2026.2`).

PR #3909, opened 2026-07-21
[SDK] Export useFormGroupOptional / FormGroupContext from the components SDK

mw-keoz:
### What

Export `useFormGroupOptional` and the `FormGroupContext` / `FormGroupProvider` symbols from the components SDK barrel (`sdk/components/index.ts`).

### Why

A Studio extension loaded as a module-federation remote currently cannot drive conditional behavior for fields **inside a field-collection row** (e.g. show `links[].url` only when that row's `linkType = external`).

Each row renders as `<Form.Group name={[index, 'data']}>`, and per-row binding works only because the host's `Form.Item` prepends the accumulated group prefix taken from `useFormGroupOptional().name`. `Form.useWatch` (antd) does **not** apply that prefix, so a plugin that watches a row-relative path reads the form root and gets `undefined`.

The clean fix — reading the host's group prefix via `useFormGroupOptional()` — is impossible from a plugin today because the hook and its context are not re-exported from the components SDK barrel, so the remote cannot obtain the host's context instance (only `react` is a shared MF singleton). `useFormGroupOptional` is already used internally (e.g. `data-object/.../data-component.tsx`); this just makes it part of the public SDK surface.

### Change

```diff
 export * from '@Pimcore/components/form/form'
 export * from '@Pimcore/components/form/form-kit'
+export * from '@Pimcore/components/form/group/provider/use-form-group-optional'
+export * from '@Pimcore/components/form/group/provider/form-group-provider'
 export * from '@Pimcore/components/form/layouts/item-spacer/item-spacer'
```

Purely additive to the public API; no runtime/behavior change for the host.

### Testing

- `npm run build-sdk` succeeds; both symbols appear in the generated SDK types.


mw-keoz:

ValeriaMaltseva (approved):
@mw-keoz
Hi! Thanks a lot for your contribution! :)

Only one thing worth mentioning: if you want to get the update sooner, please rebase it onto the 2025.4 or 2026.2 branch. Otherwise, I'll merge the PR tomorrow morning.

PR #3910, opened 2026-07-21
[Element Editor] Render element validation errors as a list without the error. prefix

mw-keoz:
### What

Render element-validation errors as a bulleted list of the server-provided violation messages, instead of one `error.`-prefixed line.

### Why

When saving an element fails validation, the backend (`studio-backend-bundle`) returns HTTP 422 with `errorKey = error_element_validation_failed` and `message` holding the human-readable, server-combined violation text (one violation per line).

`ApiErrorViewUI` treats `errorContent.errorKey` as an i18n key and renders `t(`error.${errorKey}`)`. Because the value is free text (not a translation key), i18next returns the raw string with a literal `error.` prefix, collapsed onto a single line. Users see e.g.:

> error.The button text must not be empty. Link 1: A text must be provided.

This affects **every** element whose validator fails, not just custom classes.

### Fix

Detect the `ELEMENT_VALIDATION_FAILED` case and render the message directly through the existing `SanitizeHtml` (DOMPurify keeps `ul`/`li`), bypassing `t()`: a single violation renders as a plain line, multiple as an unordered list. A small unit-tested helper (`formatValidationErrorHtml`) does the split/format.

- `api-error-view-ui.tsx` — add the `ELEMENT_VALIDATION_FAILED` branch (falls through to the existing behavior for every other error).
- `utils/format-validation-error.ts` (+ `.test.ts`) — new helper, POCL header included.

No change to any other error path.

### Testing

- `npx jest format-validation-error` — single line, multi-line → `<ul>`, blank-line/whitespace trimming, empty input → `''`.
- Manual: save an object with ≥2 failing mandatory/regex validators → the modal shows a clean bulleted list, no `error.` prefix; a single violation shows as one line.
- Content is rendered via the existing `SanitizeHtml`/DOMPurify, so only `ul`/`li` and inline text survive (no new XSS surface).

### DIFF
--- a/assets/js/src/core/modules/app/error-handler/components/api-error-view-ui.tsx
+++ b/assets/js/src/core/modules/app/error-handler/components/api-error-view-ui.tsx
@@ -14,6 +14,8 @@ import { isString, isUndefined } from 'lodash'
 import { type IErrorGetContent } from '@Pimcore/modules/app/error-handler/types'
 import { DEFAULT_ERROR_CONTENT } from '@Pimcore/modules/app/error-handler/classes/api-error'
 import { SanitizeHtml } from '@Pimcore/components/sanitize-html/sanitize-html'
+import { ErrorKeyTypes } from '@Pimcore/modules/app/error-handler/constants/errorTypes'
+import { formatValidationErrorHtml } from '@Pimcore/modules/app/error-handler/utils/format-validation-error'
 
 interface IApiErrorViewUIProps {
   errorContent: IErrorGetContent['data']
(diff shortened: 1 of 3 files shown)


ValeriaMaltseva (changes requested):
@mw-keoz 
Hi! I really appreciate your help with this. Before proceeding further with the PR, could you please create a feature task for it? This is needed to follow our development flow: Task → PR → Review → Merge.

Once this step is completed, I'll review the code in detail. Thanks! :)

ValeriaMaltseva:
@mw-keoz Hi! I'll close the PR. If you would like to proceed with it, please create an issue and then reopen the PR with the follow-up comments and changes. Thanks!

PR #3911, opened 2026-07-21
[Element Editor] Fix relation grid search not matching visible column values

mw-keoz:
### What

Fix search in the many-to-many object relation editor so it matches the visible column values, not only the object path.

### Why

When a many-to-many object relation is configured with `visibleFields`, the search box above the grid should match those visible column values. `applySearchFilter` (`core/components/many-to-many-relation/hooks/use-value.ts`) joins each grid row to its visible-field values **by id**: it builds a `Map` keyed by `field.id`, dropping any entry with a nil id (`.filter(field => !isNil(field?.id))`), then looks up `visibleFieldsMap.get(item.id)`.

But the producer (`visibleFieldsValue` in `many-to-many-object-relation.tsx`) emits only the column key/value pairs, **with no `id`**. Every entry is therefore dropped from the map, the visible-field branch never matches, and search silently falls back to matching `fullPath` only — so typing a value that is plainly visible in a column returns nothing.

### Fix

Attach the row `id` to each produced record so the join resolves.

```diff
   const visibleFieldsValue = useMemo(() => {
     return mergedGridFullData.map(item => {
-      return item?.columns?.reduce<Record<string, any>>((acc, col) => {
+      const columns = item?.columns?.reduce<Record<string, any>>((acc, col) => {
         acc[col.key!] = col.value
         return acc
       }, {})
+
+      return columns === undefined ? undefined : { ...columns, id: item.id }
     })
   }, [mergedGridFullData])
```

The `fullPath`/id/type fallback branch is untouched.

### Testing

- Manual: open an object with a many-to-many object relation that has `visibleFields` configured; assign a few objects; type a value that appears in a visible column (not in the path). Before: no match. After: the row matches. Path/id/type search unchanged.

### DIFF
--- a/assets/js/src/core/components/many-to-many-relation/hooks/use-value.test.ts
+++ b/assets/js/src/core/components/many-to-many-relation/hooks/use-value.test.ts
@@ -1,0 +1,102 @@
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+import { renderHook } from '@testing-library/react'
+import { useValue, type DisplayManyToManyRelationValue, type ManyToManyRelationValue } from './use-value'
+
+// Isolate the hook from its data/UI dependencies; the search path taken here is
+// synchronous (no valid pathFormatterConfig), so formatPath is never invoked.
+jest.mock('@Pimcore/modules/data-object/hooks/use-data-object', () => ({
+  useDataObject: () => ({ id: 1 })
+}))
+jest.mock('@Pimcore/modules/data-object/hooks/use-format-path', () => ({
+  useFormatPath: () => ({ formatPath: jest.fn(), hasUncachedItems: () => false })
+}))
+jest.mock('@Pimcore/components/modal/alert-modal/hooks/use-alert-modal', () => ({
+  useAlertModal: () => ({ warn: jest.fn() })
+}))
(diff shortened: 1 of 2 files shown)


mw-keoz:

mw-keoz:
Thanks for the review — both points addressed in `51c164cab`:

- **Regression coverage:** added `use-value.test.ts`, a `renderHook(useValue)` suite for the visible-column search behavior. A term present only in a column value matches **only** when the `visibleFieldsValue` entries carry the row `id`, and the filter falls back to `fullPath` when they don't — the exact contract this fix restores. A third case confirms the `fullPath` fallback still works.
- **Type-safe guard:** switched the producer's `columns === undefined` to lodash `isUndefined`, consistent with the rest of the file.

Locally green: `jest` (3/3), `tsc --noEmit` clean, `eslint` clean on both files.

FYI the red `build / install` check isn't related to this change — the `pull_request_target` workflow's `actions/checkout` refuses to check out fork-PR code ("pwn request" guard), so it fails the same way for any external fork PR.

ValeriaMaltseva:
@mw-keoz
Hi! Thanks for taking the time to contribute :)

At the moment, I can't reproduce this issue, either on **2025.4.9** or **2026.2.2**. Could you please double-check whether the problem still persists? If it does, would you mind providing some additional details or steps to reproduce it? That would help us investigate further. Thanks!

ValeriaMaltseva (changes requested):
@mw-keoz 
Hi! Just wanted to kindly remind you about this. Thanks! :)

https://github.com/pimcore/studio-ui-bundle/pull/3911#issuecomment-5132323576

ValeriaMaltseva:
@mw-keoz Hi! I'll close the PR. If there's still an issue, please reopen it and leave a follow-up comment. Thanks!

PR #3915, opened 2026-07-21
[Relation] Fix "Invalid column configuration" (422) when relation field has no visibleFields

klemensusec:
Relation preview grids with no configured "Visible Fields" fall back to default columns (id / fullpath / classname). These defaults only carried a key/title, so the POST to /data-objects/grid/{classId} sent columns without a type or config and the backend rejected the whole request with 422 "Invalid column configuration".

Make getDefaultVisibleFieldDefinitions() emit valid columns:
- id -> type: system.id,     group: [system], config: {}
- fullpath -> type: system.string, group: [system], config: {}
- classname -> type: system.string, group: [system], config: {}

The group: [system] also lets getDefaultSystemColumnSize() apply the correct default widths. This covers manyToManyObjectRelation, AdvancedManyToManyObjectRelation and ReverseObjectRelation, which all render through the shared ManyToManyObjectRelation component and this single hook.


klemensusec:

ValeriaMaltseva (approved):
@klemensusec 
Hi! Thanks a lot for your contribution. I'll merge the PR :)

PR #3930, opened 2026-07-23
[Bug] Class editor: allowed-classes select shows duplicate options and drops classes (folder nodes leak into options)

alexbaat:

Resolves pimcore/platform-version#239

`useClassDefinitionOptions` builds the options for the **allowed classes** select shown in the class editor settings of all relation field types (`manyToOneRelation`, `manyToManyRelation`, `manyToManyObjectRelation`, both `advanced*` variants and `reverseObjectRelation`). It fetches the class definition tree with `withGroup: true` and flattens it — but `flattenTreeNodes()` emitted the **folder (group) nodes themselves as selectable options** in addition to their children.

Since `ClassDefinitionTreeService` (studio-backend-bundle) derives group names from class name prefixes for classes without an explicit group, a folder is frequently named exactly like a real class. Example: classes `Material` and `MaterialGroup` are grouped into a folder named `Material`, so the options contain `Material` twice (folder + class) plus phantom entries for groups that are not classes at all (`Royalty`, …).

### Observed behaviour

* The allowed-classes dropdown shows **duplicate entries** (e.g. `Material` twice, `Product` twice).
* The duplicate option values / React keys break rendering of the virtualized dropdown list, so **some perfectly valid classes (e.g. `Series`) no longer show up** and cannot be selected.
* Folder/group names can be picked and saved as an "allowed class" even though no such class exists.

### Reproduction

1. Create classes `Material`, `MaterialGroup` and `Series` (no explicit group set).
2. Add a `manyToOneRelation` field to any class and open its settings in the class editor.
3. Open the *Allowed classes* select: `Material` appears twice, `Series` is missing.

### Fix

Skip folder nodes in `flattenTreeNodes()` and only recurse into their children — folders are grouping containers, not selectable class definitions. This mirrors what the neighbouring `useClassRelationFieldsOptions` hook already does. The special `folder` option added via the `includeFolder` parameter is unaffected.

Added a Jest spec covering: folder nodes excluded, no duplicate values when a folder shares its name with a class, `includeFolder` behaviour, and the not-yet-loaded state.

## Additional info

The root cause has two halves: the prefix-based auto-grouping in `ClassDefinitionTreeService::extractGroupNameFromClassName()` (studio-backend-bundle) is what makes folder names collide with class names in the first place. This PR fixes the consumer side, which resolves the user-facing bug for all relation field types; rethinking the prefix grouping itself is left out of scope.

Reproduced on v2026.1.6 and the bug is still present on `2026.x`/v2026.2.1.

### DIFF
--- a/assets/js/src/core/modules/field-definitions/dynamic-types/hooks/use-class-definition-options.spec.ts
+++ b/assets/js/src/core/modules/field-definitions/dynamic-types/hooks/use-class-definition-options.spec.ts
@@ -1,0 +1,98 @@
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+import { renderHook } from '@testing-library/react'
+import { useClassDefinitionGetTreeQuery, type ClassDefinitionTreeNodeItem, type ClassDefinitionTreeNodeFolder } from '@sdk/api/class-definition'
+import { useClassDefinitionOptions } from './use-class-definition-options'
+
+jest.mock('@sdk/api/class-definition', () => ({
+  useClassDefinitionGetTreeQuery: jest.fn()
+}))
+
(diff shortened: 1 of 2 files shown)


alexbaat:
recheck

alexbaat:

ValeriaMaltseva (approved):
@alexbaat 
Hi! Appreciate your effort on this. I've checked the changes and will merge the PR :)

Issue #256, opened 2024-05-21
Login - Forgot Password


fashxp:

mattamon:
Should we keep the original logic with sending an e-mail to login or should be switch to a more common approach to forward to a password reset form and you need to enter the password twice there?

drietsch:
I'd be in for the idea of @mattamon.

martineiber:
We will use the same approach like it the admin-classic-ui


When requested from Studio it should send a "Studio Link"
When requested from Admin Classic is should send the Classic link.

lukmzig:
As discussed with @markus-moser and @martineiber the process should be as follows:
1. POST request to recovery password endpoint:` /pimcore-studio/api/user/reset-password`
This endpoint accepts:
- username 
- resetPasswordUrl - should be FE url which will be used in step 2
Endpoint will geenerate a token and append it to the url and send an email to user
2. After creating the FE url, FE sends the token to the second POST endpoint `/pimcore-studio/api/login/token`
The endpoint authenticates the user based on the token and returns UserInformation on success.
On success FE redirects user to the my profile page or any other form where he can update his password

PR #3941, opened 2026-07-28
fix(objects): translate collapsible panel titles in BaseView #256

albertmueller:
**Fixes**: https://github.com/pimcore/platform-version/issues/256

Summary
  - BaseView only translated props.title via t() inside the CardView branch. The AccordionView branch (used whenever collapsible is true) rendered props.title raw.
  - This is visible on any collapsible panel, most notably ClassificationStore group panels in the object editor (ClassificationStoreItem), which render the group name as title and are always collapsible — so group names never went through translation and were shown as-is instead
  of using the studio translation domain.
  - Fix: compute translatedTitle once above the useMemo and use it for both the AccordionView and CardView title props, instead of duplicating the translation logic only in the CardView branch.

  Root cause
  assets/js/src/core/modules/element/dynamic-types/definitions/objects/layout-related/views/base-view.tsx:
  - AccordionView: title={ wrapWithIcon(props.title) } — no translation
  - CardView: title={ wrapWithIcon(isNonEmptyString(props.title) ? t(props.title) : props.title) } — translated

  Changes
  - base-view.tsx: introduce translatedTitle, use it in both branches, add it to the useMemo dependency array.


albertmueller:
Related to the issue: https://github.com/pimcore/platform-version/issues/256

ValeriaMaltseva (approved):

ValeriaMaltseva (changes requested):
@albertmueller 
Hi! Thanks a lot for your contribution and for taking care of this.

There is just one thing I would like to ask: could you please rebase the changes onto the **2026.2 branch**, as this is related to a bug fix? As soon as you do this, I'll merge the changes.

albertmueller:
@ValeriaMaltseva done, please have a look 😊

ValeriaMaltseva (approved):
@albertmueller 
Thanks a lot for updating the branch! I'll merge the PR.

PR #3946, opened 2026-07-29
Allow for search in sent mails

### LINKED ISSUE #321: [Studio UI] Sent mails list is not searchable / filterable
### Affected Version

2026.2

### Affected capability

Core

### Steps to reproduce

Open the sent mails dashboard.

### Actual Behavior

Contrary to the classic admin, the search field is missing, which is a regression.

### Expected Behavior

Searching or filtering the mails should be possible.

alexej-d:
Adds a search input on top of sent mails overview.

Resolves [#321](https://github.com/pimcore/platform-version/issues/321)

## Additional info
Not having a search is a regression to classic admin. We have multiple clients relying on this functionality.


pimcore-deployments:
👋 **Friendly reminder — this pull request is still a draft.**

A guardrail converted it to draft because something needed fixing, and there has been no activity since. If nothing changes, it will be **closed in 14 days**.

To keep it open, just take one of these steps:

- Address the guardrail comment above, then press **Ready for review**.
- Or leave a comment explaining that it is still in progress — any activity resets this timer.

Closing is not rejection: the branch and the discussion stay, and the PR can be reopened at any time.

jcPimcore:
Hi @alexej-d 
Thanks for providing this PR!
Could you please have a look at the open issues/suggestion please?

Thanks in advance and kind regards,

Jochen

alexej-d:
> Hi @alexej-d Thanks for providing this PR! Could you please have a look at the open issues/suggestion please?
> 
> Thanks in advance and kind regards,
> 
> Jochen

Done! I have opened another connected pull request which is a requirement for the new changes. I also added a new Controller, which lives in this bundle purely because its not needed anywhere else.

jcPimcore (changes requested):
## Verdict: Needs changes

Thanks for the follow-up commit — it correctly fixes the earlier review comment about the fabricated `hasError: false` cast by switching to a dedicated endpoint. A couple of things I'd like addressed before merging, though.

### 1. Wrong bundle for the new controller

`EmailLogSearchController` lives in `studio-ui-bundle`, but it injects `GdprManagerServiceInterface` and `EmailLogServiceInterface` from `studio-backend-bundle` — both marked `@internal` in that package. Every other Email/GDPR API controller (`Email/Controller/CollectionController`, `Gdpr/Controller/SearchDataProviderController`, etc.) lives in studio-backend-bundle itself. Depending on another package's `@internal` symbols means backend-bundle could change them without any deprecation path and silently break this bundle. Could this endpoint move into studio-backend-bundle alongside its siblings, the way `SearchDataProviderController` does?

### 2. N+1 queries instead of extending the row shape

`search()` maps every matched `GdprDataRow` through `emailLogService->getEntryDetails((int) $row->getData()['id'])` — an extra single-row DB fetch per result (up to `pageSize` per request). `SentMailProvider::findData()` already holds the fully-loaded `Log $entry` for each row when it builds the `GdprDataRow` — adding `hasError`/`error` there directly would give the same correct payload with zero extra queries, and is the smaller fix at the right boundary (also addresses point 1, since it wouldn't need a new controller reaching into `EmailLogServiceInterface` at all).

### 3. Partial-result fragility

As a consequence of point 2: if a matched entry is deleted between the search and the per-row `getEntryDetails()` call, `NotFoundException` propagates uncaught and the whole request fails, dropping all otherwise-valid results.

### 4. Test coverage

No test was added for the new controller (permission check, pagination, exception path) or for the search behavior in `email-log-container.tsx`.

### Smaller note
`emailLogSearch` is typed as `EmailLogGetCollectionApiResponse` (`items: EmailLog[]`), but the controller returns `EmailLogEntryDetail[]` (superset with `bcc`/`cc`/`error`). Harmless today, but this endpoint bypasses the generated `emails-api-slice.gen.ts` — worth reconciling once there's an OpenAPI spec for it.

### Companion PR
This is linked to `studio-backend-bundle#2004` ("Normalize entry params"), which you noted is "needed in addition to" this patch — that one's still open. Could you confirm whether it needs to land together with this PR, or is independent?

alexej-d:
> ## Verdict: Needs changes
> Thanks for the follow-up commit — it correctly fixes the earlier review comment about the fabricated `hasError: false` cast by switching to a dedicated endpoint. A couple of things I'd like addressed before merging, though.
> 
> ### 1. Wrong bundle for the new controller
> `EmailLogSearchController` lives in `studio-ui-bundle`, but it injects `GdprManagerServiceInterface` and `EmailLogServiceInterface` from `studio-backend-bundle` — both marked `@internal` in that package. Every other Email/GDPR API controller (`Email/Controller/CollectionController`, `Gdpr/Controller/SearchDataProviderController`, etc.) lives in studio-backend-bundle itself. Depending on another package's `@internal` symbols means backend-bundle could change them without any deprecation path and silently break this bundle. Could this endpoint move into studio-backend-bundle alongside its siblings, the way `SearchDataProviderController` does?
> 
> ### 2. N+1 queries instead of extending the row shape
> `search()` maps every matched `GdprDataRow` through `emailLogService->getEntryDetails((int) $row->getData()['id'])` — an extra single-row DB fetch per result (up to `pageSize` per request). `SentMailProvider::findData()` already holds the fully-loaded `Log $entry` for each row when it builds the `GdprDataRow` — adding `hasError`/`error` there directly would give the same correct payload with zero extra queries, and is the smaller fix at the right boundary (also addresses point 1, since it wouldn't need a new controller reaching into `EmailLogServiceInterface` at all).
> 
> ### 3. Partial-result fragility
> As a consequence of point 2: if a matched entry is deleted between the search and the per-row `getEntryDetails()` call, `NotFoundException` propagates uncaught and the whole request fails, dropping all otherwise-valid results.
> 
> ### 4. Test coverage
> No test was added for the new controller (permission check, pagination, exception path) or for the search behavior in `email-log-container.tsx`.
> 
> ### Smaller note
> `emailLogSearch` is typed as `EmailLogGetCollectionApiResponse` (`items: EmailLog[]`), but the controller returns `EmailLogEntryDetail[]` (superset with `bcc`/`cc`/`error`). Harmless today, but this endpoint bypasses the generated `emails-api-slice.gen.ts` — worth reconciling once there's an OpenAPI spec for it.
> 
> ### Companion PR
> This is linked to `studio-backend-bundle#2004` ("Normalize entry params"), which you noted is "needed in addition to" this patch — that one's still open. Could you confirm whether it needs to land together with this PR, or is independent?

I have adjusted my code in both bundles @jcPimcore

alexej-d:
Anything else I should change @jcPimcore?

jcPimcore:
Hi @alexej-d !

Looks good! Please have a look at the small fix left for the other PR.
When all is done, there should be no blocker left and we could merge both.

Thx and kind regards,

Jochen

jcPimcore (approved):

jcPimcore:
Hi @alexej-d 
Tested and merged. Will be part of the 2026.3 release.
Thanks a lot for this great contribution!

Kind regards,
Jochen

PR #3949, opened 2026-07-29
Fix: keep Perspectives field visible for Superadmin users

jcPimcore:
## Summary
- `CustomisationAccordion` hid the Perspectives (and Roles) fields entirely whenever `isAdmin`
  was true, so once a user with a previously assigned Perspective was promoted to Superadmin,
  there was no way to view or clear that stale assignment through the Studio user-edit UI.
- Fix: the Perspectives field now always renders, regardless of `isAdmin`. The Roles field keeps
  its existing admin-only visibility unchanged — that's a separate design choice, out of scope
  for this bug.
- Companion fix in `studio-backend-bundle` (same branch name) stops the backend from restricting
  a Superadmin's allowed/active perspectives based on a stale stored assignment.

Refs PEES-1390, pimcore/platform-version#253

## Test plan
- No existing unit/RTL test convention was found for these accordion components in this repo, so
  no automated test was added here.
- [x] Manual: open a Superadmin user's Settings tab in the Studio user-edit UI and confirm the
      Perspectives field is now visible (and editable) rather than hidden.
- [x] Combined with the backend PR, confirm a Superadmin with a stale Perspective assignment gets
      the full tree/menu view instead of the restricted one.


pimcore-deployments:

PR #3958, opened 2026-08-03
Lay out listing toolbar slots in a Flex instead of a Space

alexbaat:
Fixes pimcore/platform-version#284

### Changes in this pull request

The asset and data object listing toolbars wrap their `SlotRenderer` in a `Space`:

```tsx
<Space size='small'>
  <SlotRenderer slot={ componentConfig.dataObject.listing.toolbar.left.name } />
</Space>
```

`Space` wraps each of its **direct** children in an `.ant-space-item`, and its only direct child here is the single `SlotRenderer` element. `SlotRenderer` then renders a fragment with all registered slot components *inside that one item div*.

So the `Space` never actually lays out the slot components:

* they end up as inline-level siblings in a single block container, so they are **baseline-aligned instead of centered**;
* the configured `size='small'` **gap is not applied** between them.

For the default case (one slot component) this is invisible, but as soon as a second component is registered to the slot — for instance an export button next to the core batch-actions group — it renders a few pixels off vertically and flush against its neighbour, because the two inline-flex boxes have different heights.

Using `Flex` makes every slot component a real flex child, which centers them and applies the gap. `Flex`'s `gap='small'` resolves to the same `token.sizeSM` as `Space`'s `size='small'`, so the default layout is unchanged. This also matches how the asset and data object *editor* toolbars already render their slots (`Flex` around the `SlotRenderer`).

### Measured before / after

A bundle registers one extra component (a `DropdownButton`) to `dataObject.listing.toolbar.left`, next to the core batch actions. Both buttons are 32px high. `getBoundingClientRect()` in the running app, same window for both runs:

| | before | after |
|---|---|---|
| vertical offset between the two buttons | `4px` | `0px` |
| horizontal gap between them | `0px` | `12px` (`token.sizeSM`) |

Before, the registered button's nearest slot-level ancestor computes to `display: inline-flex` inside the shared `.ant-space-item`; after, it is blockified to `display: flex` as a real flex item of the toolbar container.

### Additional info

Verified in a Studio 2026.1 project, with and without an active row selection (the left toolbar swaps the class-definition select for the selection total, i.e. a different content height) — the two buttons are pixel-identical in both states. Listings without an extra slot component are visually unchanged. `npx eslint` and `npx tsc --noEmit` are clean on this branch.


ValeriaMaltseva (approved):
@alexbaat 
Hi! Thanks for your contribution. I will merge the PR :)

PR #3959, opened 2026-08-04
[Object List] Add option to convert a column to an advanced column

### LINKED ISSUE #3421: [Advanced columns] Add conversion of  normal column to advanced column
It would be nice to have the possibility to convert a normal column into an advanced one:

Image

In the original design there was a toggle planned for that. But unfortunately that can only work reliable in one direction.
So I guess we should introduce something like a icon that allows this type of converstion?

vin0401:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/3421

## What

Grid config rows now offer converting a normal column into an advanced one, seeded with a source field pointing at the original column.

The reverse direction (advanced → normal) is intentionally omitted: an advanced column can hold several source fields, a transformer chain and static text, so collapsing it back is lossy with no safe rule.

## How

Convertibility is owned by the pipeline **source field dynamic types**, not by the grid config UI. `DynamicTypePipelineAbstract` gains one method:

```ts
getConversionConfig (column: AvailableColumn, config: Record<string, any>): Record<string, any> | undefined
```

It returns the seed config for a single pipeline item, or `undefined` when the type cannot represent the column. Defaulting to `undefined` keeps `staticText` and all transformers out of it without any change to them.

This sits next to the existing `isAvailableForSelection()`, so the convert affordance and what the pipeline offers for selection are decided by the same classes and cannot drift apart. It also means a bundle can extend conversion by registering a new source field type or overriding an existing one — no UI change needed.

Per type:

- **`simpleField`** — claims a column whenever the backend offers its key as a source field, carrying `groupId`/`keyId` over for classification store columns so the converted column resolves the same value. Object bricks work unchanged, as the collector builds the identical `field.BrickType.name` key format.
- **`relationField`** — seeds *both* the relation and a field of the related object. Only seeding the relation is not viable: `RelationFieldConfig::$field` is a required non-nullable string and `Column::getAdvancedColumnConfig()` reads it without a null coalesce, so a half-seeded item would `TypeError` as soon as the grid or preview requests data. Classification store fields are skipped when picking the default, as their group/key cannot be inferred. A relation whose allowed classes share no field stays unconvertible rather than producing a config the backend rejects.

The helper assembling the column keeps `locale` and `width`, seeds the title from the column label, and leaves `__meta.uniqueId` empty so the row remounts with its pipeline form expanded. Titles are made unique per locale, since an advanced column is identified by its title at runtime and columns are matched to their data by key + locale.


pimcore-deployments:

**Added (1):**
- `listing.grid-config.convert-to-advanced`

Issue #889, opened 2025-01-17
[Data Object Editor] Many to one relation - display mode "Inline Search"


markus-moser:
![Image](https://github.com/user-attachments/assets/50e86592-97f9-4516-bf35-1cf14550cb6b)

PR #3960, opened 2026-08-04
[Data Object Editor] Many to one relation - display mode "Inline search"

### LINKED ISSUE #287: Data Object Editor: many-to-one relation ignores display mode "Inline search"
### Affected Version

2026.1 (verified on `pimcore/studio-ui-bundle` 2026.1.6; the code path is unchanged on `2026.x`)

### Affected capability

Data Objects

### Steps to reproduce

1. Add a **Many-To-One Relation** field to a data object class and restrict it to a single class.
2. In the field settings set **Display mode** to **"Inline search"** and save the class.
3. Re-open the class definition — the setting is stored correctly (`'displayMode' => 'combo'` in `var/classes/definition_<Class>.php`).
4. Open an object of that class in the Data Object editor.

### Actual Behavior

The field renders the path reference input, exactly as with "Display path reference". There is no way to search; you can only pick the related object through the element selector.

The value is stored and delivered correctly — `displayMode` is a public property on `Pimcore\Model\DataObject\ClassDefinition\Data\Relations\AbstractRelations` and the layout is passed through raw — but the object data component never reads it: `DynamicTypeObjectDataManyToOneRelation.getObjectDataComponent()` always renders `ManyToOneRelation`. Only `manyToManyObjectRelation` implements a `combo` branch (labelled "Tag field" there).

So the Display mode dropdown is effectively dead UI for `manyToOneRelation`: both options render the same widget.

### Expected Behavior

"Inline search" renders a searchable field for the related class, as the option promises and as the equivalent setting does for many-to-many object relations.

---

A pull request implementing this is already open: pimcore/studio-ui-bundle#3960. This issue is filed for traceability, per the issue-link requirement on that repository.

Supersedes pimcore/studio-ui-bundle#889, which describes the same gap.

alexbaat:
Resolves pimcore/platform-version#287

Supersedes #889, which tracked the same gap before tracking issues moved to `platform-version`.

## Problem

The class definition editor offers **Display mode → "Inline search"** on a `manyToOneRelation` and stores it correctly (`displayMode: 'combo'`, persisted via `AbstractRelations::$displayMode`). The object data component never reads it, though: `DynamicTypeObjectDataManyToOneRelation.getObjectDataComponent()` always renders `ManyToOneRelation`, the path reference input. Only `manyToManyObjectRelation` implements a combo branch (labelled "Tag field" there).

So picking "Inline search" changes nothing anywhere — both options render the same widget.

## Change

Adds a single-select inline search field for `manyToOneRelation`, used when `displayMode === 'combo'` **and** the relation is restricted to objects of exactly one class.

The combo offers its options through a search on a single data object class, so it can only represent a relation restricted to exactly that. A relation that also accepts assets, documents or object folders keeps the path reference input, which can address every element type — otherwise those targets would become unselectable.

New files, next to the owning dynamic type and mirroring the existing many-to-many combo:

- `components/many-to-one-relation/components/combo-field/many-to-one-relation-combo-field.tsx`
- `components/many-to-one-relation/components/combo-field/hooks/use-combo-field-data.ts` — paged full-text search plus label resolution
- `components/many-to-one-relation/components/combo-field/hooks/use-relation-source-object-id.ts`

### Notes on three details that are easy to get wrong

**Labels.** A `manyToOneRelation` has no `visibleFields`, so option labels come from the field's `pathFormatterClass` (`data-objects/format-path`), falling back to the object path when none is configured. `getAvailableDataObjectColumnConfigurationForRelation` cannot be reused here — it throws for anything that is not a (Advanced)ManyToManyObjectRelation — hence a separate hook rather than reusing the many-to-many `useComboFieldData`.

**Source object.** `format-path` resolves the field definition from a concrete object. That object is taken from the row being edited (grid cell edit modal, via `getGridCellEditComponent`), the first selected row (batch edit), or the element context (object editor). Inside a listing the element context is the *folder* being listed, and the endpoint answers `404 Object with ID: x not found` for it, so it is only used outside a listing.

**Searching.** Filtering happens server side (`filterOption={false}`). Filtering the options client side only matches the rendered labels, so typing an object's name finds nothing whenever those labels are object paths.

`useClassDefinitionCollectionQuery` maps the allowed class name to its id rather than `useClassDefinitions()`, because that context is folder-scoped inside a listing (`ClassDefinitionsProvider` filters its items by `elementId`) and would not know the related class.

## Tests

- `dynamic-type-object-data-many-to-one-relation.test.ts` — the rendering decision: combo vs. path input per display mode, the object-only guard with a case per mixed configuration (assets, documents, object folders), and that the grid cell editor forwards the edited row id.
- `use-combo-field-data.test.ts` — class id resolution, the debounced full-text term, page reset on search, paging on scroll versus a fully loaded list, and label resolution through the path formatter including both fallbacks.

`npm run check-types` clean, `npm run lint` clean for the touched files, `npm test` 265 passed.

## Manual verification

Verified against a project with ~8.000 objects in the related class and a `pathFormatterClass` configured, in all three places the field renders — object editor, grid cell edit modal and batch edit — covering searching, paging, selecting, saving and clearing.

## Review feedback addressed

- Inline search no longer claims mixed asset/document/folder relations; regression cases added per type.
- The `inherited` state is forwarded to `Select`, so combo-mode fields keep the inherited styling.
- The component moved next to the owning dynamic type and is no longer exported through the SDK barrel.
- Added hook coverage for search, pagination and label resolution.


ValeriaMaltseva (changes requested):
Hi @alexbaat, and thanks a lot for your contribution! The PR already looks good.

One thing I think we need before merging: the class definition editor is not updated. **field-definition-many-to-one-relation-form-fields.tsx**:151-158 still offers Display mode → "Inline search" for every configuration, including the ones usesInlineSearch() now rejects - more than one allowed class, assets or documents allowed, folder-only. There the setting saves and then silently does nothing, which is the dead-UI symptom pimcore/platform-version#287 describes, just moved to a different configuration.

Could you mirror the usesInlineSearch() condition there and disable the "Inline search" option when it does not apply, with the reason in the Form.Item tooltip? The form currently watches only displayMode, so it would also need objectsAllowed, assetsAllowed, documentsAllowed and classes. field-definition-many-to-many-object-relation-form-fields.tsx:25,44-56 does the same kind of reconciliation for visibleFields if you want a reference.

Please keep the option in the list and disabled rather than removing it - with a stored displayMode: 'combo' antd would otherwise render the raw value as the label. No need to migrate stored values.

alexbaat:
Hi @ValeriaMaltseva, thanks for the review — good catch, the editor side was
indeed still offering the mode unconditionally. Pushed in 48635b0.

**Shared condition.** The eligibility check moved out of
`DynamicTypeObjectDataManyToOneRelation` into `supportsInlineSearch()` in
`helpers/relations/inline-search.ts`, without the `displayMode` part. The data
component is now `props.displayMode === 'combo' && supportsInlineSearch(props)`
and the editor calls the same function, so the two cannot drift apart. It takes
`Partial<IRelationAllowedTypesClassDefinition>` and compares explicitly against
`true`, because the form values are `undefined` until their switch is touched.

**Editor.** `field-definition-many-to-one-relation-form-fields.tsx` now watches
`objectsAllowed`, `assetsAllowed`, `documentsAllowed` and `classes` next to
`displayMode`. The "Inline search" option gets `disabled` when the relation is
not eligible, and the `Form.Item` gets a tooltip naming the reason:

> Inline search is only available when the relation allows objects of exactly
> one class. A relation that also allows assets, documents or object folders
> uses the path reference instead.

Added to all seven locale files. The tooltip is only set while the option is
unavailable — happy to make it permanent if you would rather have the
constraint documented at all times.

**No reconciliation.** I deliberately did not mirror the `visibleFields` reset
from `field-definition-many-to-many-object-relation-form-fields.tsx`: a stored
`displayMode: 'combo'` is left as it is, and the option stays in the list so
antd keeps rendering its label. That follows your last paragraph — say the word
if you would prefer the value to fall back to `grid` once the configuration
stops supporting it.

**Tests.** `inline-search.test.ts` covers the predicate itself: a single allowed
class, more than one class, folder-only, assets or documents allowed, objects
disallowed, empty and undefined `classes`, and the untouched switches.
`field-definition-many-to-one-relation-form-fields.test.tsx` covers the wiring:
the option is disabled and the tooltip is set for each rejected configuration,
neither for the supported one, and the option is still present with a stored
`combo`. Full suite 280 passed, `check-types` and lint clean.

ValeriaMaltseva (approved):
Hi @alexbaat! Thanks for the really quick update to the PR, and once again, thanks for your contribution. I'll merge it ;)

PR #3961, opened 2026-08-04
[Batch Delete] Use the batch delete-info endpoint in the grid delete confirmation

### LINKED ISSUE #3666: [Batch Delete Modal] Accordion for listing the items
Image

xIrusux:
Resolves #3666

The grid batch-delete confirm dialog fired one `delete-info` request per selected row (N parallel requests) to decide between "Delete" and "Delete permanently", and only opened once all of them resolved — large selections hit browser connection limits and froze the UI without any loading indicator.

It now uses the aggregated endpoint from pimcore/studio-backend-bundle#1954 (released in 2026.2.1, already covered by the composer constraint `^2026.2.1`):

- A single `POST /elements/{elementType}/batch-delete-info` request runs first; the confirm dialog then opens once, fully formed — the same fetch-first shape as the folder delete flow, no post-open `update()` patching.
- `canUseRecycleBin` decides the OK label as before.
- The aggregated `hasDependencies` now drives the dependency warning: a concrete "These items have child objects or other dependencies" line shows only when dependencies actually exist, and no line when there are none. If the pre-check request fails, the API error is raised via the central error handler and the dialog closes (no under-warning "Delete" state); the now-unused generic `element.delete.batch.dependencies-warning` key was removed from all languages.
- The batch-info call is a read-only pre-check, so the enhanced slice drops its `Elements` cache invalidation (same pattern as `elementLock`/`elementUnlock`); the request handle is `reset()` afterwards, which also fixes the previously never-unsubscribed per-id cache entries.
- OpenAPI snapshot: only the new `batch-delete-info` path was spliced in (taken from a freshly generated spec); SDK regenerated via `build-api-client`.
- New translation key `element.delete.batch.dependencies-warning.confirmed`, translated for en/de/fr/it/es/sv/no.

## Additional info
- The wording of the new dependency warning line could use a quick look from @idaiv — the previously agreed static text ("There may be dependencies, delete anyway?") intentionally remains as the error fallback, where the uncertainty is real.
- Verified locally against a demo instance: the endpoint responds and the rebuilt frontend is served; eslint, the element-module jest suites, and a full `build-app` are green.


pimcore-deployments:

**Added (1):**
- `element.delete.batch.dependencies-warning.confirmed`

**Removed (1):**
- `element.delete.batch.dependencies-warning`

Issue #3922, opened 2026-07-22
[DataObject Editor] Field Container "width" setting is ignored (always renders full width)


alexbaat:
## Bug description

The **width** configured on a *Field Container* layout element in a DataObject class definition is ignored by the Studio UI. Regardless of the value entered in the class editor, the container is always rendered at full width and its child fields are distributed with equal flex.

The `width` value is correctly persisted in the class definition (e.g. `'width' => 400` on the `Layout\Fieldcontainer` node), but it is never read or applied when rendering the object editor.

## Steps to reproduce

1. Open a DataObject class in the class editor.
2. Add a **Field Container** (layout `hbox` or `vbox`) with a few fields inside it.
3. Set the Field Container **width** to e.g. `400`.
4. Save the class and open/edit an object of that class in Studio.

## Expected behaviour

The Field Container respects the configured width (as it did in the legacy admin UI).

## Actual behaviour

The width setting has no effect. The container always takes full width and the children are laid out with equal `flex: 1`.

## Cause

In `assets/js/src/core/modules/element/dynamic-types/definitions/objects/layout-related/components/field-container/field-container.tsx` (tag `v2026.2.1`):

- Line ~23 — the component destructures only `{ children, collapsible, collapsed, noteditable, layout = 'hbox', fieldLabel }`; `width` is not among the props it reads.
- Line ~30 — the outer container is hardcoded to `className='w-full'`.
- Line ~38 — each child is wrapped in `<Flex flex={ 1 } />`, forcing equal distribution.

So the persisted `width` is never mapped to a style/width on the container.

## Affected versions

Reproduced on `2026.1.6`; source inspection confirms the same behaviour on the current `2026.2.1`. The only 2026 commits touching this file (#3580 vbox support, #3754 noteditable flag) do not address width.

## Environment

- Pimcore Studio UI bundle: 2026.1.6 (also confirmed in 2026.2.1 source)
- Pimcore: 2026.1
- PHP 8.5, Symfony 7.4

PR #3962, opened 2026-08-05
[DataObject Editor] Apply the configured width and height to a Field Container

alexbaat:

Fixes pimcore/platform-version#292
pimcore/studio-ui-bundle#3922

`FieldContainer` never read the `width` and `height` that the class editor lets you set
on a layout element. It destructured only `children`, `collapsible`, `collapsed`,
`noteditable`, `layout` and `fieldLabel`, and hardcoded the outer `Flex` to
`className='w-full'` — so whatever was persisted on the `Layout\Fieldcontainer` node was
dropped and the container always spanned the full width.

Both values now run through the existing `toCssDimension()` helper and are applied to the
container.

## Additional info

- **`0` counts as "not set".** `Layout::$width` and `Layout::$height` default to `0`, and
  `toCssDimension()` already maps `0`, `''` and `null` to `undefined`. So an
  unconfigured container renders exactly as before and only an explicitly configured
  value changes anything — the fix is additive. Numbers become pixels; strings such as
  `'50%'` pass through unchanged.
- **`w-full` is applied only when no width is configured.** It resolves to
  `width: 100%`, so an inline width would win on specificity anyway, but keeping the two
  mutually exclusive avoids two competing width declarations on the same element.
- **`height` is included on purpose.** The class editor offers width *and* height for
  every layout element (`field-defintion-layout-form-fields.tsx`, with `width-tooltip` /
  `height-tooltip`) and both were equally ignored here, so fixing only one would leave
  the other as a follow-up issue. Both are opt-in for the reason above.
- **The children keep `flex={ 1 }`.** Their even distribution is unrelated to the
  container's persisted width, and changing it would alter the rendering of every
  existing field container. Left untouched deliberately, and covered by a regression
  assertion in the test.
- **Related, not touched:** `Panel`, `Fieldset`, `Accordion` and `Tabpanel` ignore their
  configured width/height as well; only `Region` reads it, and only from its children (as
  `maxWidth`). That looked like a bigger, separate change, so it is out of scope here —
  noted in pimcore/platform-version#292 so it does not get lost.

### Tests

New `field-container.test.tsx` covering: unset width keeps `w-full`, the class-definition
default of `0` is treated as unset, a numeric width becomes `px`, a string width passes
through, height is applied independently, the `vbox` layout and the children's `flex: 1`
are unchanged, and the width still lands on the container when a `fieldLabel` wraps it in
a `Form.Item`. Four of the seven fail without the change.

The real `Flex`/`Form` and the `@sdk/utils` barrel pull in untranspiled ESM
(`antd-style`) that jest cannot load, so they are mocked — the same approach as
`keyed-list.test.tsx`, with the two helpers taken from their actual modules via
`jest.requireActual`.

`npm run lint`, `npm run check-types` and `npm run test` (19 suites / 242 tests) all pass.


ValeriaMaltseva (approved):
@alexbaat Hi! Thanks a lot for your contribution! I'll merge the PR :)

PR #3964, opened 2026-08-06
[Telemetry] Drain the telemetry outbox after login

martineiber:
Part of https://github.com/pimcore/product-management/issues/1243

Browser-side half of server-side product telemetry. Companion to pimcore/pimcore#19336 and pimcore/studio-backend-bundle#1980 — **both must merge first** (this calls the outbox endpoints the backend PR adds).

## Why

Instances behind a DMZ cannot reach the relay from the server, so their telemetry would sit in the outbox forever. This registers a post-login app loader that drains it from the browser instead: fetch an encrypted batch from the Studio backend, forward it to the relay, then acknowledge it.

## Safety

The browser is only a courier — the batch is encrypted with the instance product key, so the page cannot read the telemetry it carries.

A batch is acknowledged only on a **confirmed relay acceptance**, never on a bare 2xx, so a captive portal or proxy returning 200 cannot cause data loss. Any other outcome leaves the batch in the outbox for the next login or for the maintenance job. The loop is bounded per session and every failure path exits quietly — telemetry must never disrupt a login.

No CSP change is needed: `https://license.pimcore.com/` is already in the bundle's default `connect-src` allow-list.

## Notes

- Three new files under `assets/js/src/core/modules/telemetry/` plus one import line in `bootstrap.ts`.
- Headless module — no UI surface.


xIrusux (approved):
LGTM :)

PR #3965, opened 2026-08-06
chore: fix MAJOR Sonar code smells (S4782, S9011, S1854)

ValeriaMaltseva:
**Fixes**: https://github.com/pimcore/platform-version/issues/299

Clears **31 of the 71** open MAJOR issues reported for this project on [SonarCloud's new-code period](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&severities=MAJOR&sinceLeakPeriod=true&id=pimcore_studio-ui-bundle). Mechanical cleanups only — no feature or API changes.

### S4782 — redundant `?` plus `undefined`

Props declared as `foo?: SomeProps['bar']` where the referenced prop is itself optional, so the indexed access type already includes `undefined`.

Wrapped in `Exclude<…, undefined>`, which leaves the resulting prop type **provably identical** while removing the redundancy:

```ts
- padding?: BoxProps['padding']
+ padding?: Exclude<BoxProps['padding'], undefined>
```

Two alternatives were deliberately rejected:

- **`NonNullable<…>`** — also strips `null`. Several of these resolve to `boolean | null | undefined` (the `noteditable` props), where `null` is meaningful. `tsc` caught this.
- **Dropping the `?`** — would turn optional props into required ones and break every caller.

Since a number of these are SDK-public component props, keeping the type identical matters for external bundles.

### S9011 — `<button>` without an explicit `type` (6 issues)

Added `type="button"` in `main-nav` (2), `hotspot-image-container` (2), `tree-element-item` and `send-test-email-button`. All are plain click handlers, so no behavior changes — but the implicit `submit` default is a hazard should any of them ever be rendered inside a form.

### S1854 — useless assignments (5 of 6 issues)

- Dropped two unused `t` bindings and their `useTranslation` imports (`use-recycle-bin`, `search-replace-assignments-provider`).
- Dropped the unused `pageSize` destructure in `translation-merger`. The `Pagination` wrapper deliberately `Omit`s that prop and owns page size internally via `defaultPageSize: 20`, which already matches the hook's default — so there was nothing to wire up.
- Removed the write-only `jobTitle` state and its effect from the CSV and XLSX export modals, along with the `useElementDraft` call that existed only to feed it. `CsvDownloadJob`/`XlsxDownloadJob` take their title from a static translation key and accept no per-instance title, so nothing consumed it.

### Not addressed

Left out because they need more than a mechanical fix:

- **10× S8786** (regex backtracking) — rewriting risks silently changing match behavior. 6 are in build/bundler tooling that runs at build time on trusted input, so the ReDoS angle is low; some may be *Won't fix* candidates.
- **2× S6819** (`role="button"` on non-button elements, in `default-cell` and `tree-expander`) — converting to real `<button>`s touches grid keyboard/click semantics and styling.
- **1× S7721** (move `getModalComponent` to outer scope).

### Verification

Run locally on this branch, all passing:

| Check | Result |
|---|---|
| `npm run check-types` | pass |
| `npm run lint` | pass |
| `npm test` | pass — 235 tests, 18 suites |

No regression test accompanies this PR: every change is either type-level or dead-code removal, with no behavior to pin down. The existing suite plus `tsc` is what guards it. CI is the final validator.


ValeriaMaltseva:
Tracking issue: pimcore/platform-version#299

That issue covers all 71 MAJOR SonarCloud findings on the new-code period. This PR clears 57 of them; the issue also records the one false positive to dismiss (`handleOrder`) and the 13 left for separate, non-mechanical work (regex backtracking, `role="button"` accessibility, and `getModalComponent` scope).

PR #3970, opened 2026-08-07
fix: apply custom attribute key for data objects and documents

robertSt7:
## Summary
- `getElementKey()` in `element-helper.ts` only ever returned the element's own `key`/`filename`, ignoring `customAttributes.key`.
- The `icon` and `tooltip` custom attributes were already respected via `customAttributes`, but `key` was not, so setting a custom key from a `DataObjectEvent`/`DataObjectDetailEvent` listener had no visible effect in the tree, detail view, rename, or delete flows.
- `getElementKey()` now prefers `customAttributes.key` when set, falling back to the existing behavior otherwise — mirroring how `determineElementIcon()` already handles `customAttributes.icon`.

resolves https://github.com/pimcore/platform-version/issues/161

## Test plan
- [x] Added `element-helper.test.ts` covering `getElementKey()` for data objects and documents, both with and without a custom attribute key set.
- [x] `jest js/src/core/modules/element` — all 22 tests pass.
- [x] `eslint` on the changed files — clean.

Co-Authored-By: Claude <noreply@anthropic.com>


assets/js/src/core/modules/asset/services/asset-opening-service.ts
```diff
@@ -81,7 +81,7 @@ export class AssetOpeningService {
     const widgetId = getWidgetId('asset', id)
 
     return {
-      name: data?.filename,
+      name: getElementKey(data, 'asset'),
```
ValeriaMaltseva: Please use the `elementTypes.asset`

assets/js/src/core/modules/data-object/services/data-object-opening-service.ts
```diff
@@ -85,7 +85,7 @@ export class DataObjectOpeningService {
     const widgetId = getWidgetId('data-object', id)
 
     return {
-      name: data?.key,
+      name: getElementKey(data as Element, 'data-object'),
```
ValeriaMaltseva: Please use the `elementTypes.dataObject`

assets/js/src/core/modules/document/services/document-opening-service.ts
```diff
@@ -82,7 +82,7 @@ export class DocumentOpeningService {
     const widgetId = getWidgetId('document', id)
 
     return {
-      name: data?.key,
+      name: getElementKey(data as Element, 'document'),
```
ValeriaMaltseva: Please use the `elementTypes.document`

ValeriaMaltseva (changes requested):
The changes already look very good! I just left a few really small comments. Please take a look, and then it's LGTM from my side ;)

ValeriaMaltseva (approved):
Looks good, let's merge it.

PR #3971, opened 2026-08-07
[DataObject] [Classification Store] Fix selection modal: grid does not use full modal width and columns are not resizable

albertmueller:
### Description
  
  In the ClassificationStore selection modal (opened from the DataObject edit view when adding
  a Collection/Group/Key to a `classificationstore` field), the selection grid in all three tabs
  ("Collection", "Gruppe"/"Group", "Nach Schlüssel gruppieren"/"Group by key") renders at a fixed
  width far narrower than the modal itself, and its columns cannot be resized by dragging the
  column border — unlike virtually every other grid in Studio (e.g. Predefined Properties,
  Translations, Document Types).
  
Screenshot 2026-08-07 085247
  
  ### Steps to reproduce
  
  1. Open any DataObject with a `classificationstore` field in Studio.
  2. Go to the field's "Attribute" tab and click "Hinzufügen" / "Add".
  3. Observe the selection modal: the grid only fills roughly the left third of the modal width,
     with a lot of empty space to the right.
  4. Try to drag the border between two column headers to resize a column — nothing happens.
  
  ### Expected behavior
  
  - The grid should fill the full width of the modal (as it does with `autoWidth` elsewhere).
  - Columns should be resizable via drag, consistent with other Studio grids.
  
  ### Actual behavior
  
  - The grid renders at the sum of TanStack's default column widths (150px each), leaving most
    of the modal empty. 
  - Columns are not resizable.
  
  ### Root cause
  
  `ClassificationStoreDataTab` and `ClassificationStoreCallbackTab`
  (`src/core/modules/element/dynamic-types/definitions/objects/data-related/components/classification-store/components/classification-store-modal/components/classification-store-data-tab/`)
  render their `<Grid>` without the `resizable` and `autoWidth` props. Both props are required for
  `Grid` to enable column resizing (`Resizer`) and full-width layout — see
  `src/core/components/grid/grid.tsx`. Every other grid in the codebase that wants full-width /
  resizable behavior sets both props explicitly (e.g.
  `src/core/modules/predefined-properties/table/table.tsx`).
  
  ### Environment
  
  - pimcore/studio-ui-bundle 2026.2.x
  - Pimcore 2026
  
  Fixes https://github.com/pimcore/platform-version/issues/304
  
  A fix is ready: adds `resizable`/`autoWidth` to the two `<Grid>` usages and sets explicit column
  sizes with one `autoWidth` column per tab (Collection/Group/GroupByKey), so the last column fills
  the remaining space. PR will follow from branch `fix/classification-store-modal-grid-width`.
  
Fixed state:
Screenshot 2026-08-07 093606

Fixes pimcore/platform-version#304


kingjia90 (approved):
Verified against the `Grid` component API and all call sites:

- `autoWidth` / `resizable` are real `Grid` props (`grid.tsx:60`, `grid.tsx:250`) and the per-column `meta: { autoWidth: true }` fill-pattern is the same one used by other Studio grids (e.g. saved-searches tab) — this brings the CS modal in line with the rest of Studio.
- The one-autoWidth-column-per-grid validation in `grid.tsx` (`validate if only one column has autoWidth`) is respected: each tab defines exactly one flexible column.
- The two shared renderers (`classification-store-data-tab.tsx`, `classification-store-callback-tab.tsx`) are only consumed by the three updated tabs (collection / group / group-by-key), so no other grid picks up the new behaviour unintentionally.
- `useMemo` on the column defs with `[t]` is correct and now actually load-bearing, since the grid's autoWidth validation memoizes on column identity.
- Copilot: no findings. CI green (build, lint, types, Codeception, PHPStan, Sonar).

I added `Fixes pimcore/platform-version#304` to the description so the issue-link guardrail passes and the issue auto-closes on merge.

ValeriaMaltseva (approved):
@albertmueller
Hi! I really appreciate your help. Thanks! :)
I'll merge the PR.

PR #3974, opened 2026-08-07
fix: element tree renders empty when the root node has not loaded yet

robertSt7:
## Summary

- The element tree's always-visible child list rendered the still-loading root node with a non-null assertion, producing a node without an `id` and a `tree-list-undefined` test id that could never resolve its own children.
- Falls back to the already-known node id while the root node is loading, and compares root/tree ids as strings so a numeric-vs-string id mismatch can no longer hide an already-loaded root.

Fixes https://github.com/pimcore/platform-version/issues/149

## Test plan

- [x] Added a regression test in `element-tree.test.tsx` that reproduces the `tree-list-undefined` rendering with the root node not yet loaded, and asserts the tree list keys off the real node id instead.
- [x] Added a companion test confirming the root node still renders as its own node once loaded.
- [x] `npx jest ./js/src` — all 21 suites / 249 tests pass.
- [x] `tsc --noEmit` — no errors.
- [x] `eslint` on the changed files — no errors.


ValeriaMaltseva (approved):
LGTM :)

PR #3975, opened 2026-08-07
fix: preserve explicitly cleared classification store values on save

robertSt7:
## Summary

- `filterInheritedFields()` (used when building the classification store payload on save) filtered fields by checking `value !== null` instead of using the `isInherited()` predicate it was given, and its group/language "emptiness" check used the same null-based logic.
- Clearing every populated key in a group/language (e.g. resetting a quantity value's amount *and* unit together) made every remaining field null, which collapsed that group to `[]` in the outgoing payload.
- The backend's `ClassificationStoreAdapter::setStoreValues()` iterates each group's data to apply changes; an empty array has nothing to iterate, so the group is silently skipped and the previous value is left in the database — the field appears cleared in the UI but reverts on reload.
- Fixed `filterInheritedFields()` to actually use `isInherited()` to decide what to omit, and to derive group/language emptiness from the filtered result instead of re-scanning raw values for `null`. Explicit clears (including a fully-cleared quantity value) are now preserved and sent to the backend; fields that are genuinely still inherited continue to be omitted.

Related: https://github.com/pimcore/service-operations/issues/981

## Test plan

- [x] Added regression tests in `group-value.test.ts` covering: an explicit clear is preserved, a mixed cleared/populated group keeps both, a genuinely inherited field is still omitted, and a group where every field is inherited still collapses to `[]`.
- [x] `npx jest .../classification-store/utils/group-value.test.ts` — all 4 tests pass (previously all 4 failed against the old implementation).
- [x] `npx eslint` on the changed files — clean.
- [x] Reproduced the underlying bug against a live Pimcore Studio instance via the REST API before the fix (an empty-array group payload left the stored value untouched) and confirmed normal saves are unaffected.


ValeriaMaltseva (approved):
Looks good, let's merge it :)

PR #3976, opened 2026-08-10
[Application Logger]: Add possibility to sort the table and all extensions

lukmzig:
Part of https://github.com/pimcore/data-importer/issues/562

## Additional info
Adapt sorting approach so data hub extensions are able to re-use it for the logger tabs


assets/js/src/core/modules/application-logger/components/table/table.tsx
```diff
@@ -78,7 +79,9 @@ export const Table = ({ items, isLoading, sorting, onSortingChange }: TableProps
     columnHelper.accessor('translatedPriority', {
       header: t('application-logger.columns.type'),
       size: 60,
-      enableSorting: true
+      // Filterable via the sidebar instead: the cell shows a translated label, so any database
+      // side ordering would follow the enum or the untranslated value, not what is on screen.
+      enableSorting: false
```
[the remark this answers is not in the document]
lukmzig: It was decided that priority sorting is not necessary and provided a complicated case with ENUMs. For priority filtering its enough.

assets/js/src/core/modules/application-logger/utils/sorting.ts
```diff
@@ -0,0 +1,51 @@
...
+  component: 'component',
+  source: 'source'
+}
+
+export const DEFAULT_APPLICATION_LOGGER_SORT_FILTER: SortFilter = {
+  key: 'id',
+  direction: 'DESC'
+}
+
+export function mapSortingToSortFilter (sorting: SortingState): SortFilter {
```
xIrusux:
```suggestion
export const mapSortingToSortFilter = (sorting: SortingState): SortFilter => {
```

we try to use es6 where we can i.e. see above

assets/js/src/core/modules/application-logger/utils/sorting.test.ts
```diff
@@ -0,0 +1,58 @@
+/**
```
xIrusux: we dont very often add tests in the frontend. I will start a conversation about this in standup tmr to align as a team.

When we test we do it mainly on: pure functions or a bug that we fix needing extra documentation.

So pushing it is fine by me.

xIrusux (approved):
LGTM, just minor comments

assets/js/src/core/modules/application-logger/utils/sorting.test.ts
lukmzig: Test is removed now :)

assets/js/src/core/modules/application-logger/utils/sorting.ts
lukmzig: applied by https://github.com/pimcore/studio-ui-bundle/pull/3976/commits/6614a9edd121d6b06c39e3f7f8d117c628f84a43

PR #4000, opened 2026-08-20
[Document Editor] Inline upload for relation editable

### LINKED ISSUE #1620: [Document Editor] Inline upload for relation editable
Same as https://github.com/pimcore/studio-ui-bundle/issues/1043 for document editables

ValeriaMaltseva:
Closes #1620

Adds inline upload to the document `relation` editable, mirroring what the sibling `relations` (many-to-many) editable already offered.

> [!NOTE]
> This PR was reworked after #4039 was merged. #4039 already added the upload button to `ManyToOneRelation` (shown whenever assets are allowed and the field is enabled), the `assetUploadPath` prop, the post-upload subtype guard and its `asset-upload-type-not-allowed` message. This PR keeps that merged behavior unchanged and only adds what is still missing for #1620.

## What changed

**`ManyToOneRelation`** gained a `disableInlineUpload` opt-out prop that hides the upload button #4039 introduced. This mirrors the existing `disableInlineUpload` prop on the sibling `ManyToManyRelation` component, so both relation components are now configured the same way.

**The document `relation` editable** wires up the template config — same option names as the `relations` and `image` editables, so all of them behave alike:

```twig
{{ pimcore_relation('myRelation', {
    'uploadPath': '/content',
    'disableInlineUpload': false
}) }}
```

`uploadPath` was already declared in the editable's config type but unused; it is passed to the component as `assetUploadPath` now. `disableInlineUpload` also covers the request in #1966 for the many-to-one case. The upload button itself appears on the editable via #4039's default (assets allowed + enabled) — no opt-in needed.

No new upload machinery: it reuses the shared upload modal, so duplicate-name conflict handling, size limits, the upload queue and the iframe forwarding (`element.openUploadModal`) all apply unchanged.

## Also fixed: subtype was not enforced on upload in many-to-many

#4039 fixed this for `ManyToOneRelation`; the same gap existed in the **many-to-many** toolbar. `allowedAssetTypes` was enforced for the element selector (`createElementSelectorConfig`) and for drag & drop (`dndIsValidData` → `isAllowedSubType`), but **not** for uploads. A relation restricted to `image` therefore accepted an uploaded PDF — a value neither of the other two paths could have produced.

The toolbar now filters uploaded assets through `isAllowedSubType` before assigning, and warns when some were dropped. The check runs *after* upload rather than as an `accept` filter on the file dialog, because the server decides the asset type from the MIME type, while `accept` is only a dialog hint. The assets are already created at that point, so the message says so instead of pretending they were rejected.

## Translations

- `relations.upload.subtype-not-allowed` (new) — phrased count-independently, since the many-to-many upload can partially fail; added to all seven locales at the English position.
- `asset-upload-type-not-allowed` (from #4039, used by `ManyToOneRelation`) — was English-only; the six missing locale translations are synced now (partly by the automated translation-sync workflow on this branch).

`Asset` is left untranslated per the glossary.

## Verification

- `tsc --noEmit` — clean
- `eslint` on all touched files — clean
- Full jest suite — **602 tests passing**, including #4039's existing `ManyToOneRelation` upload tests (unchanged and still green), a new test for `disableInlineUpload`, and 3 tests for the many-to-many partial-rejection behaviour
- Storybook story added under `Components/Data Entry/Relation/ManyToOneRelation` → *Configurations* → **With Inline Upload**

**Not verified in a browser.** My workspace has a `2026.2` backend against this `2026.x` frontend, and `POST /assets/exists/{id}` (the batched upload exists-check from #3986) only exists on backend `2026.x` — so every upload 405s locally, independent of this change. The iframe → parent `openUploadModal` hop and the button's visual placement are therefore verified as code paths only, not empirically. CI plus a review on an aligned environment should cover the rest.

## Note for the reviewer

The many-to-many fix is a pre-existing bug in a sibling component, not part of #1620. It is here because it is the same defect #4039 fixed for many-to-one, and the same guard. Say the word if you would rather have it as its own PR.


pimcore-deployments:

**Added (2):**
- `relations.upload.subtype-not-allowed`
- `many-to-one-relation.upload.subtype-not-allowed`

Issue #354, opened 2024-06-18
[Versions] Show date/time with seconds


markus-moser:
![image](https://github.com/pimcore/studio-ui-bundle/assets/4639428/3eeda616-207a-472c-bcab-ed26e8571c13)

vs. in the admin UI

![image](https://github.com/pimcore/studio-ui-bundle/assets/4639428/84951f74-7bb5-4bad-843d-c0bbd7b389fd)

In the design we only have the minutes but the seconds are important.

We also need to show the time in addition to the date on the left column:

![image](https://github.com/pimcore/studio-ui-bundle/assets/4639428/edaaccf0-0515-46b4-ae35-4ed8f83be777)

There we also need the time with seconds. If we get some space problems there we would need to think a bit about the design but it is very important to show both as otherwise users get confused when the item get's saved multiple times in a short period of time.

Related to https://github.com/pimcore/studio-ui-bundle/pull/349

markus-moser:
Blocked by #365

PR #4012, opened 2026-08-21
[Translations] Fix: search & field filters in Translations sidebar don't trigger on Enter/search icon

albertmueller:
Summary

- The top search field in the Translations "Search & Filter" sidebar only wrote the typed value into a draft store on change; clicking the search icon or pressing Enter (which antd's Input.Search reports via onSearch, not onChange) did nothing. Fixed by wiring onSearch to apply the term immediately.
- Field filters (e.g. the "German" text column) used a plain Input with no search icon, only committing on blur into the draft store — the value was lost unless "Apply" was clicked afterwards. Fixed by switching to SearchInput, adding an optional onCommit callback threaded through
  DynamicFilterProvider → FieldFilters → the translations field-filter hook, so Enter/search-icon immediately applies the filter, and stretched the input to full sidebar width.

Changes

- filters/filters.tsx: search term control now applies on onSearch, not just onChange.
- components/dynamic-filter/provider/dynamic-filter-provider.tsx: added commit()/onCommit alongside the existing setData()/onChange.
- components/field-filters/field-filters.tsx: added an onCommit prop, mirroring the existing onChange handling.
- dynamic-type-field-filter-text-component.tsx: switched from Input to SearchInput, wired onSearch to commit(), set full width.
- translations/filters/hooks/use-translations-field-filter-editor.ts: added onFilterCommit, which writes the committed field filter directly into both the draft and applied filter stores.
- translations-sidebar/components/filter-tab/filter-tab.tsx: wired the new onCommit handler into FieldFilters.

Fixed state:
image

Fixes https://github.com/pimcore/platform-version/issues/354
The issue was also reported as an enterprise issue: PEES-1392


ValeriaMaltseva (approved):
@albertmueller 

Thanks a lot for the report and the fix. I also pushed commits to your branch to change one thing: the search control inside the field filter rows.

**Why.** `DynamicTypeFieldFilterTextComponent` is not specific to Translations - it powers the text field filter in six sidebars. Switching it to SearchInput added a magnifier icon and a clear button to all six. There were two issues with that:

1.  Only Translations wires onCommit. In the other five cases, commit() falls back to onChange, which simply writes the draft value. As a result, the button looked actionable while doing no more than leaving the field already did. A control that promises an action it cannot actually perform is worse than no control.
2. Visually it reads as a second search box competing with the sidebar's own search field right above it, once per filter row. It doesn't match how the rest of the filter rows look.

**What it does now.** The component keeps the plain antd `Input` it always used and reacts to `onPressEnter` instead: Enter applies immediately where the host exposes `commit()`, and behaves like leaving the field where it doesn't. The rendered markup is therefore unchanged in every host - no styling or selector shifts - and any sidebar can opt into immediate-apply later by passing `onCommit`.

One consequence worth stating plainly: the "search icon and full sidebar width" part of #354's expected behaviour is intentionally not implemented. Enter and the top search field cover the functional half; the per-row icon we're deliberately leaving out.

I hope this explains the reasoning behind the changes.

I will merge the PR soon.

Issue #291, opened 2024-05-21
Data Object Editor - Data Type Inheritance


fashxp:
- [x] show inherited data
- [x] link to data origin
- [ ] restore changes

![image](https://github.com/pimcore/studio-ui-bundle/assets/8792145/eb0a6e76-ed48-4fc1-b278-bc639c82979b)

https://www.figma.com/design/kTxowT29eCHwNh5NspDGGu/Pimcore-Components?node-id=1182-99023&t=TH6LzmPptxSxTMON-4

markus-moser:
Follow-up issues:

- #1045
- #1046
- #1047

PR #4026, opened 2026-08-25
[Data Object Editor] Restore (inherited) changes

### LINKED ISSUE #1045: [Data Object Editor] Restore (inherited) changes
See #291

ValeriaMaltseva:
Resolves #1045, the remaining unchecked item of #291.

Adds a per-field **Restore** action that undoes the change which broke a field's inheritance, so the field takes its value from the origin object again.

## Scope

`inheritanceData` carries only the origin id and a flag per field, never a value, so the value a field *would* inherit is known only for fields that were inherited when the editor was opened. The action is therefore offered for those, and not for fields that already carried an own value — covering the latter needs a backend endpoint that returns the inherited value plus origin id, which does not exist today.

Field collections, calculated values, consent, URL slugs and reverse relations never get the action, because `supportsInheritance()` is false for them.

## How it works

`useRestoreInheritance` picks the branch by where the value actually lives.

**Plain fields** are held by the Ant form store:

1. `form.resetFields([name])` puts the loaded (inherited) value back. Ant does not report a reset through `onValuesChange`, so the field is not counted as changed again.
2. The field is separately persisted as empty, because an auto save may already have written the own value into the draft and the draft is cumulative on the server. The empty value comes from the field type (`DynamicTypeObjectDataAbstract.getEmptyValue`) — `null` for almost everything, but `[]` for blocks, whose adapter takes `array` and would throw a `TypeError` on `null`. I checked every data adapter; `BlockAdapter` is the only one that cannot take `null`.
3. The inheritance state goes back to the entry it was loaded with.

**Object bricks and the classification store** hold their values in a `Form.KeyedList`, which the Ant form store never sees, and each encodes inherited fields in its own payload — `null` for bricks, absence for the store. So the owner of the list restores the field instead: it drops its own "field was changed" bookkeeping and re-emits the payload, and the displayed value fixes itself because `getMergedValue` then reads the field from the loaded data. A synchronous marker (`restoredFieldsRef`, mirroring the existing `changedFieldsRef`) is needed because restoring the inheritance state only schedules a React update, which the payload built right after must not wait for.

`KeyedList` gains one optional `onFieldRestore` prop, the counterpart of the existing `onFieldChange`. It is the channel the label uses to reach the owner. `components/pipeline/*` passes nothing and is unaffected.

## Blocks

Block inheritance is all-or-nothing at field level — `BlockAdapter` does not implement `DataInheritanceInterface`, so the generic path yields one state for the whole block field. Restore therefore sits on the block's own label, not on its inner fields, which matches that granularity.

## Drive-by fix

`getMergedValue` dropped the inherited values of a classification store group once that group held no own values any more. `filterInheritedFields` deliberately encodes such a group as an empty list, `getFieldList` reports that as a field of its own, and the merge then wrote over the values already merged in for the keys of the group. Restoring the last own key of a group made this reachable. `getMergedValue` had no tests; it has three now.

## Verified

- `tsc --noEmit`, `eslint`, and the jest suite (562 tests) pass; `npm run dev-sdk` builds.
- New tests: the break → restore cycle in the inheritance state provider incl. a localized path, the payload shape (nested locales, siblings kept, no mutation), both branches of `useRestoreInheritance`, and the `getMergedValue` regression.
- **Not verified in a browser.** Worth a look at: a plain field, a localized field (only the shown locale must change), a block (the save must not fail), an object brick, and a classification store key — each restored and then the tab reloaded, which is what proves the field was persisted as empty.

## Follow-ups, not in here

- Restoring a field whose own value was already persisted before the editor was opened (needs the backend endpoint above).
- A store-level Restore for a whole classification store field: inheritance there is tracked per group × language × key, and reverting the field would also have to undo added and removed groups (`activeGroups`, `groupCollectionMapping`, `DELETED` markers). Different semantics, and destructive enough to want a confirmation.
- Two latent issues found while reading, deliberately left alone: `getMergedValue` in `group-value.ts` throws if `originalValue` holds a group that the current value does not (the sibling `brick-value.ts` guards this); and `DataComponent` builds `virtualFieldName` without the `Form.List` path, so a field inside a block can pick up the inheritance overlay of a same-named top-level field.


pimcore-deployments:

**Added (2):**
- `inheritance-active-unknown-origin`
- `inheritance-restore`


20 validation error(s) predating this PR were present at the merge base. This run also fixed 4 of them; 16 still remain. The remainder is not blocking — later runs keep chipping at it.

PR #4042, opened 2026-08-28
Add configurable pagination options and default page size

### LINKED ISSUE #306: [Studio Ui][Pagination] More / Custom pagination options
### Affected capability

Other

### Feature description

Within Admin UI it was possible to set a custom amout of items shown per page. In Studio UI this is limited to 10/20/50/100 per Page. Can we get the custom option in Studio UI back, or maybe at least an "Show all" entry. This would be pretty helpful.

Image
Image

SimonCremer:
Introduces a new `pimcore_studio_ui.pagination` configuration node with two options:

- `page_size_options`: a comma separated list of page sizes offered in the page-size changer (default `10,20,50,100`), replacing the hardcoded `[10, 20, 50, 100]` used across all listing pagination components.
- `default_page_size`: the page size a listing starts with before the user picks another option (default `20`), now used by the user references tab and the main element listing paging provider instead of a hardcoded `20`.

Both values are wired from Configuration.php through PimcoreStudioUiExtension/services.yaml into DefaultController's appConfig, and consumed on the frontend via the appConfig singleton (app-config.ts).

Also documents both settings under
doc/03_Configuration_and_Administration/01_Configuration/07_Pagination.md, including a note that the largest `page_size_options` value must not exceed Elasticsearch/OpenSearch's `index.max_result_window` result limit.

Resolves pimcore/platform-version#306


xIrusux:
@SimonCremer Thank you for this PR. An adapted version of it has been merged and will be available with the next platform release.
See: https://github.com/pimcore/studio-ui-bundle/pull/4109

PR #4050, opened 2026-08-31
[Studio UI] Add aria-expanded to tree nodes

idaiv:
## Summary
- Expandable tree nodes now expose `aria-expanded={true|false}` so screen readers announce folder open/closed state
- Only set on nodes with children; leaf nodes omit the attribute
- `aria-selected` deferred to ARIA roles PR (requires `role="treeitem"`, current role is `"button"`)

## Test plan
- [ ] Inspect a folder node in DevTools — verify `aria-expanded="false"` when collapsed
- [ ] Expand the folder — verify `aria-expanded="true"`
- [ ] Inspect a leaf node — verify no `aria-expanded` attribute
- [ ] Test with screen reader (NVDA/VoiceOver) — verify folder state is announced


idaiv:
Superseded by #4048 which includes aria-expanded along with the full tree navigation rewrite.

PR #4051, opened 2026-08-31
[Studio UI] Add Home/End key navigation to tree

idaiv:
## Summary
- Home key jumps to the first visible tree node, End to the last
- Consolidated navigation functions into a single `gotoNode` helper
- Standard WAI-ARIA tree keyboard pattern

## Test plan
- [ ] Focus any tree node
- [ ] Press Home — verify focus moves to the first visible node
- [ ] Press End — verify focus moves to the last visible node
- [ ] Verify existing arrow up/down still works


idaiv:
Superseded by #4048 which includes Home/End keys along with the full tree navigation rewrite.

PR #4052, opened 2026-08-31
[Studio UI] Let UsersRolesDropdown offer the signed-in user

fashxp:

Adds an opt-in `includeCurrentUser` prop to `UsersRolesDropdown`. Default `false`, so every existing consumer behaves exactly as before.

```ts
includeCurrentUser?: boolean
```

## Why

The component filters the signed-in user out of the selectable options:

```ts
userList?.items?.filter(item => item.id !== userData?.id)
```

That is correct for its original use case, sharing an entity with *other* people, where offering yourself is meaningless.

It is wrong when the same component is reused to edit an **access-control list**. There, adding yourself is usually the very first thing someone does, and today it is impossible through the UI. Existing self-entries still render and can be removed, so the gap is add-only, which makes it easy to miss in review and confusing in use.

Found while building an OAuth access allow-list in `data-hub-simple-rest` (see pimcore/data-hub-simple-rest#356), where an administrator enabling OAuth cannot grant it to themselves.

## Why a prop, rather than fixing it in the consumer

There is no consumer-side workaround. The exclusion is not in the API or the backend:

- `UsersShareController::getUsersShareList()` returns every user and excludes nobody.
- `UserGetShareCollectionApiArg` is `void` — the endpoint takes no parameters.
- The filter compares against `useUser()` resolved *inside* the component, so it drops the signed-in user whatever `userList` a caller passes.

The only alternative would be for each such consumer to rebuild the dropdown on the same two queries.

## Additional info

Labels are unaffected. `userNameById` is built from the full `userList` and already falls back to the current user's own name, so a self-entry rendered a proper label before this change too; the prop only governs which options are offered.

Verified with `tsc --noEmit` and `eslint` on the changed file, both clean, and exercised in a browser through the `data-hub-simple-rest` allow-list: with the prop set the signed-in user appears in the Users tab and can be selected, and the default-`false` behaviour is unchanged everywhere else.


xIrusux (approved):
LGTM :)

PR #4062, opened 2026-09-01
Fix notification popups surviving session-expiry redirect to login

### LINKED ISSUE #3743: [Running Jobs] Notifications bar appears on log out screen
It appears as a bug to still have internal data when the user has logged out. 

Image

sholzer:
## Summary
- Fixes: #3743 
- Running-jobs and user-to-user notification popups render into a single global AntD notification context mounted above the router, so they weren't destroyed when a 401 (expired session) redirects to the login page — they stayed visible on the login screen.
- Added an unmount cleanup (`notificationApi.destroy(...)`) to both notification components so the popup is torn down when the owning component unmounts.

## Repro (before fix)
1. Log into Studio in two browsers as different users.
2. Send a notification from browser A to browser B's user (or trigger a running job).
3. See the popup in browser B.
4. Delete all cookies in browser B.
5. Trigger any request in browser B (e.g. expand the element tree) → redirected to login.
6. Notification popup is still visible on the login screen.

## Test plan
- [x] `eslint` clean on both changed files
- [x] `jest` passes for `execution-engine` (76/76) and `notifications` (52/52) modules
- [x] Manually verified end-to-end: sent a self-targeted notification via a role/group recipient (direct self-recipient is dropped server-side), confirmed popup appears, then expired the session and confirmed the popup no longer persists on the login screen (after rebuilding frontend assets)
- [x] Manual two-browser, two-user verification of the running-jobs notification path

Co-Authored-By: Claude <noreply@anthropic.com>


xIrusux (approved):
LGMT :)

PR #4065, opened 2026-09-02
[Studio UI] Add WAI-ARIA toolbar pattern with roving arrow key focus

idaiv:
## Summary
- Toolbar component now implements WAI-ARIA toolbar pattern
- `role="toolbar"` on the Flex container
- ArrowLeft/Right moves focus between focusable children with wrapping
- Reduces Tab stops — keyboard users Tab into toolbar once, then use arrows

## Test plan
- [ ] Focus a toolbar button via Tab
- [ ] Press ArrowRight — focus moves to next button
- [ ] Press ArrowLeft — focus moves to previous button
- [ ] At last button, ArrowRight wraps to first
- [ ] Tab exits the toolbar to the next focusable element

Relates to https://github.com/pimcore/product-management/issues/372


idaiv:
Closing — the generic Toolbar component wraps arbitrary children (dropdowns, split buttons) that need their own keyboard behavior. Adding role="toolbar" + arrow key interception at this level breaks Enter/Space on dropdowns and steals ArrowLeft/Right from menus. Needs a more targeted, opt-in approach in a future PR.

Issue #3207, opened 2026-03-24
[Translate] - all Bundles into main languages


xIrusux:
Bundles: 

- [x]  quill-bundle	 https://github.com/pimcore/quill-bundle/pull/70
- [x]  asset-metadata-class-definitions https://github.com/pimcore/asset-metadata-class-definitions/pull/368
- [x]  frontend-permission-toolkit https://github.com/pimcore/frontend-permission-toolkit/pull/109
- [x]  statistics-explorer https://github.com/pimcore/statistics-explorer/pull/175
- [x]  object-merger https://github.com/pimcore/object-merger/pull/86/files
- [x]  workflow-automation-integration-bundle https://github.com/pimcore/workflow-automation-integration-bundle/pull/110
- [x]  direct-edit https://github.com/pimcore/direct-edit/pull/102
- [x]  data-quality-management-bundle https://github.com/pimcore/data-quality-management-bundle/pull/270/files
- [x]  headless-documents 	https://github.com/pimcore/headless-documents/pull/288
- [x]  openid-connect https://github.com/pimcore/openid-connect/pull/154
- [x]  copilot-bundle https://github.com/pimcore/copilot-bundle/pull/817
- [x]  translations-provider-interfaces https://github.com/pimcore/translations-provider-interfaces/pull/237	
- [x]  studio-backend-bundle https://github.com/pimcore/studio-backend-bundle/pull/1732
- [x]  ee-web-to-print-bundle https://github.com/pimcore/ee-web-to-print-bundle/pull/93	
- [x]  portal-engine https://github.com/pimcore/portal-engine/pull/883
- [x]  ee-ecommerce-framework-bundle	https://github.com/pimcore/ee-ecommerce-framework-bundle/pull/133/files
- [x]  data-hub https://github.com/pimcore/data-hub/pull/1068
- [x]  ee-personalization-bundle	https://github.com/pimcore/ee-personalization-bundle/pull/79
- [x]  backend-power-tools-bundle https://github.com/pimcore/backend-power-tools-bundle/pull/635	
- [x]  workflow-designer https://github.com/pimcore/workflow-designer/pull/184
- [x]  ee-customer-data-framework https://github.com/pimcore/ee-customer-data-framework/pull/82
- [x]  data-importer https://github.com/pimcore/data-importer/pull/576
- [x]  studio-ui-bundle https://github.com/pimcore/studio-ui-bundle/pull/3236

Languages:  

[...]

PR #4070, opened 2026-09-02
3207 - [Studio UI] - Translations (nl, pl)

xIrusux:
## What

Adds full Dutch (`studio.nl.yaml`) and Polish (`studio.pl.yaml`) Studio UI translations — 2829 keys each, mirroring `studio.en.yaml` line for line.

Part of the nl/pl rollout across all Studio bundles (tracking: #3207). `nl`/`pl` are deliberately **not** yet added to the translation skill's `data/languages.yaml` — that flip happens last, once all bundles carry the files, so the CI translation bot doesn't start backfilling onto unrelated PRs.

## How

- **Seeded** from human-provided nl/pl translation exports covering 2674 keys.
- **Conformance pass** against the i18n skill's glossary pins and guidelines:
  - `Asset`, `Object Brick`, `Field Collection`, `Custom Layout`, `Layout`, `Workspace` kept untranslated (matching de/fr/it/es/sv/no).
  - nl: fixed `Aanmakdatum` → `Aanmaakdatum` (8×), `Geerfde` → `Geërfde`, and the Alert/Warning log-severity collision (`priority-level.2` → `Alarm`).
  - pl: CRLF → LF; restored three placeholder values damaged by unquoted-number YAML parsing (`00:00:00.000`, `00:00:10.000`, `25`).
- **Translated the 157 keys** missing from the exports (newer en additions: saved-search, ownership-management, edit-lock, areablock, notification settings, batch delete, …).
- Dropped 3 stale export keys no longer present in en.

## Verification

`validate_translations.py --dir translations --languages nl,pl`: no missing/extra keys, order matches en, placeholders/quoting/plurals clean. The only 2 errors per file are the unquoted `yes:`/`no:` keys — identical in **every** existing locale (rooted in en itself); fixing that class is a separate en-wide change.

Known limitation: Polish plural forms are restricted to `_one`/`_other` to keep key parity with en (no `_few`/`_many`), same as all existing locales.

Closes pimcore/product-management#1338

### DIFF
--- a/translations/studio.nl.yaml
+++ b/translations/studio.nl.yaml
@@ -1,0 +1,2871 @@
+common.owner: Eigenaar
+common.modification-date: Wijzigingsdatum
+common.globally: Globaal
+common.apply-selection: Selectie toepassen
+common.clear: Wissen
+cache.clear-cache-success: Cache succesvol gewist
+cache.clear-full-page-cache-success: Volledige paginacache succesvol gewist
+cache.clear-temporary-files-success: Tijdelijke bestanden succesvol gewist
+dont-ask-again: Niet meer vragen
+widget.missing-context.title: Ontbrekende context
+widget.missing-context.description: Geen gerelateerd element gevonden
+widget.missing-tab-context.description: Deze functie wordt niet ondersteund voor het gerelateerde element
+block: Blok
+error.error_something_generic_went_wrong: Er is iets misgegaan
+error.error_permission_denied: Je hebt geen toestemming om deze actie op dit element uit te voeren.
+error.error_element_exists: Element bestaat al
+error.error_folder_exists: Map bestaat al
+error.error_invalid_argument: Ongeldig argument
+error.error_widget_name_missing: Widgetnaam moet worden opgegeven
+error.error_widget_name_invalid: Widgetnaam is ongeldig (alleen alfanumerieke tekens, underscores en spaties zijn toegestaan, minimaal 3 en maximaal 80 tekens lang)
+error.error_validation_failed: Validatie mislukt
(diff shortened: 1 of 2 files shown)


pimcore-deployments:

#### ⚠️ Ambiguous keys (review carefully)

# Translation Sync Notes

## `yes` / `no` keys — pre-existing English defect, not fully fixable here

`studio.en.yaml` defines these keys unquoted:

```yaml
yes: Yes
no: No
```

Per the guidelines, `yes`/`no` are boolean words in YAML and must be quoted as keys, so each
`studio.{locale}.yaml` (de, fr, it, es, sv, no) had the same unquoted-key defect, reported by the
validator as "non-string key (needs quoting)".

**Chosen reading:** quoted the keys in all six target files (`'yes': ...`, `'no': ...`), per the
YAML quoting rules in the guidelines (§4, boolean/null words must be quoted).

**Effect / alternative:** because `studio.en.yaml` itself is unquoted, PyYAML parses its `yes`/`no`
keys as booleans (`True`/`False`), so they are dropped from the English key set the validator
compares against. Quoting the target keys as strings now makes the validator report them as
"extra key not in English" instead of "non-string key" — the error count per file is unchanged
(2 errors), just recategorized. A full fix requires quoting `yes:` / `no:` in `studio.en.yaml`
itself, which is out of scope for this run (English is not editable here). Left as-is rather than
reverting, since quoted string keys are the technically correct form on the target side; a
follow-up PR should quote the same two keys in `studio.en.yaml`.

#### ℹ️ Pre-existing translation errors (not from this PR)

12 validation error(s) predating this PR are still present. They are not blocking it; clearing them needs another run or a dedicated backfill.

PR #4071, opened 2026-09-02
[Search] Close the search when opening an element from a full path tag

### LINKED ISSUE #1879: [Search] clicking full path on data object search opens image in the background
**Steps to reproduce**

- Login as admin
- Open the search
- Switch to data objects
- Search for 'Allfa'
- Click on the full path in the search result of a data object

**Actual behaviour**
- Image opens in the background

Image

**Expected behaviour**
- Data object should be opened and search should be closed

**Environment**

- Studio Nightly
- Windows
- Google Chrome

sholzer:
Resolves #1879

Clicking the full path in a search result opened the element but left the search modal in front of it, so the element appeared to open in the background. A double click then also hit the grid's row handler, which is what closed the modal.

ElementTag's own onClick called openElement without letting the surrounding UI react, while every sibling open path — the three open-element decorators and the general tab's result item — pairs openElement with close(). The fix covers all three typed tabs (Documents, Assets, Data Objects) plus any other element tag rendered inside the modal.

ElementTag is exported from the SDK, so it does not reference the search module. It fires onElementClick() from a feature-neutral ElementClickBehaviorContext that defaults to a no-op, and the three search listings supply close() as that callback — keeping the close logic in the search listing layer, next to the OpenElementDecorator that already closes the search on row double click. The callback fires after the awaited openElement().

## Additional info

**Steps to reproduce**
- Open the search, switch to Data Objects, search for a term
- Click the full path in a result row
- Before: the object opens behind the still-open modal; a double click was needed to close it
- After: the object opens and the search closes on a single click

**Verified**
- `npx eslint` on both changed directories — clean
- `npx tsc --noEmit` — clean
- Existing frontend suite — 57 suites / 417 tests pass


xIrusux (commented):
More of an idea to check out to keep topics close together rather than a definite change request :)

Not so long ago I implemented select & close on doubleclick (for lisist in many to many relation fields) 

Thats why I am thinking the tag click could live alongside it in the OpenElementDecorator 

i.e. 
`They could both live in `OpenElementDecorator`. Decorators can wrap the `ContextComponent` slot with a provider, the same way `withRowSelectionContext` or `withPagingContext` do. So next to `with-open-element-event.ts` there could be a `with-open-element-context.tsx`:

```tsx
export const withOpenElementContext = (Component: AbstractDecoratorProps['ContextComponent']): AbstractDecoratorProps['ContextComponent'] => {
  const OpenElementContextComponent: AbstractDecoratorProps['ContextComponent'] = () => {
    const { close } = useSearch()

    return (
      <ElementClickBehaviorProvider onElementClick={ close }>
        <Component />
      </ElementClickBehaviorProvider>
    )
  }

  return OpenElementContextComponent
}`

That would mean:

- the three listing files stay untouched
- one decorator owns both "open from search" behaviours
- any listing that composes OpenElementDecorator gets the tag close for free

xIrusux (approved):

PR #4078, opened 2026-09-03
Feature/element selector context 431

### LINKED ISSUE #431: Context information in element selecor
### Affected capability

Data Objects

### Feature description

When selecting an element within the old Admin UI, there was some context information available, like the field name. Depending on that context, you could do custom things when searching and selecting elements for a relational field. This context is missing in the Studio UI.

There have been some discussions and requests from me and at least two other customers who want that functionality in a similar way in the Studio UI. Unfortunately, no one from Pimcore has participated in those threads.

One proposed solution from a customer was to fully replace the UI element selector and implement custom logic in the backend, but I don't think fully replacing the standard UI components is a good idea.

I checked the code and think almost everything is in place to add at least some context to the element selector. With minimal effort, it would be possible to add the field name as a custom filter (similar to tags). On the backend, anyone who wants custom behavior for specific fields could implement a FilterInterface for their specific needs.

On the backend side, only a minimal change is necessary: adding a new value, e.g., system.fieldName, to ColumnType.

On the UI side, only small changes are needed, which shouldn't have a big impact. It would be possible to create a decorator very similar to TagFilterDecorator to add the field name to column filters. Adding the field name is even easier, since the value is already available when the selector is created.

If you're interested in the feature, I could provide an implementation. If you prefer a different approach, let me know.

SimonCremer:
Resolves pimcore/platform-version#431

## Additional info
Add field-name filter to data-object element selector


pimcore-deployments:

PR #4087, opened 2026-09-08
[Documents] Do not send link editables as internal without an element…

### LINKED ISSUE #459: [Bug]: Link editable is sent as internal without an element behind it, breaking document rendering and saving
### Affected Version

2026.2

### Affected capability

Documents

### Steps to reproduce

1. Place a link editable on a document and point it at an internal document.
2. Delete the target document.
3. Open the document in the Studio document editor and save it, so the link value is round-tripped through the editable's transformValueForApi().
4. Render the document, or save it again.

### Actual Behavior

 Rendering and saving both fail with "An exception has been thrown during the rendering of a template".

 The link is stored as `internal: true` with `internalId: null` and, in most cases, no `internalType` key at all.

### Expected Behavior

 A link editable should only be marked as internal when an element is actually behind it.

codingioanniskrikos:
Resolves pimcore/platform-version#459
 
 A link editable was sent to the backend as `internal: true` even when no element was behind it, and `internalType` was sent as `undefined`, which `JSON.stringify` drops. This happens with a link whose target document was deleted: it still has `linktype: 'internal'`, but no id and no type left.

The backend then trusts that flag and reads `internalType` / `internalId` without checking them, which breaks rendering and saving of the whole document.

Now `internal` is only claimed when an id is present. Otherwise the value is sent as a direct link, and `internalType` is sent as `null` instead of `undefined`.

### DIFF
--- a/assets/js/src/core/modules/element/dynamic-types/definitions/document/editable/types/dynamic-type-document-editable-link.test.ts
+++ b/assets/js/src/core/modules/element/dynamic-types/definitions/document/editable/types/dynamic-type-document-editable-link.test.ts
@@ -65,6 +65,21 @@ describe('DynamicTypeDocumentEditableLink.transformValueForApi', () => {
     expect(result.fullPath).toBe('/en/some-page')
   })
 
+  // The case reported in #459: the target document was deleted, so the value still
+  // claims linktype: 'internal' but has neither an id nor a type left behind it.
+  it('sends a link whose internal target was deleted as a direct link', () => {
+    const result = transform(linkValue({ internal: null, internalType: null }))
+
+    expect(result.linktype).toBe('direct')
+    expect(result.internal).toBe(false)
+    expect(result.internalId).toBeNull()
+    expect(result.internalType).toBeNull()
+    expect('internalType' in result).toBe(true)
+    // The path the link used to resolve to is kept as a plain direct link.
+    expect(result.path).toBe('/en/some-page')
+    expect(result.fullPath).toBe('/en/some-page')
+  })
+
   it('downgrades to a direct link when the internal type is missing', () => {
     const result = transform(linkValue({ internal: 42, internalType: null }))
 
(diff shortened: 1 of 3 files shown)


pimcore-deployments:
🚫 **CI & merge-readiness guardrail failed — this PR has been converted to draft.**

A PR can only be merged when it has no conflicts with the base branch and all CI checks pass.

- CI check not successful: `cla-workflow / CLAAssistant` (failure).

pimcore-deployments:

codingioanniskrikos:

ValeriaMaltseva (changes requested):
@codingioanniskrikos 
Hi! Thanks for this - the diagnosis is spot on, and the guard sits in the right place. It
matches the invariant the rest of the feature already assumes: `use-link-data-type.ts:57`
gates `openLink` on exactly the same two fields, and `link-value-converter.ts:53-54` always
sets `internal` and `internalType` together. Sending `internalType: null` instead of
`undefined` is the load-bearing part — with the key present-but-null,
`Link.php:306` (`setDataFromEditmode`) short-circuits cleanly and the backend's own
normalisation runs instead of reading a missing key.

Three things before this can go in:

1. **Base branch.** The linked issue reports *Affected Version 2026.2*, but this targets
   `2026.x`. Bugfixes land on the active release line and forward-merge up, so please
   retarget to `2026.2` — otherwise 2026.2 users don't get the fix.

2. **Regression test.** `transformValueForApi` is a pure function and there are jest tests
   throughout this tree already (e.g.
   `document/editable/components/renderlet-editable/renderlet-content.test.tsx`), so this is
   cheap. Four cases: valid internal stays internal; internal + id but no `internalType` →
   direct; internal + `internalType` but no id → direct; direct unchanged. Please also
   assert `'internalType' in result` — the `undefined` key being dropped by `JSON.stringify`
   is the exact mechanism of this bug, and a plain deep-equal won't catch a regression.

3. **Changelog.** `DocumentLinkEditableValue` is re-exported through the public SDK
   (`js/src/sdk/modules/element/index.ts:169`), so widening `internalType` to `string | null`
   is a public type change — external bundles assigning it to `string | undefined` will now
   fail to compile. The widening is the right call (narrowing it is what caused the bug), it
   just shouldn't land silently.

Thanks in advance! :)

Issue #3371, opened 2026-04-21
[Form Kit] UI Improvements


idaiv:
The form kit implementation has inconsistency in terms of the vertical gaps between the different input fields. Would it be possible to introduce a gap rule of 8px? 

Example of the inconsistency - the top fields have the gap, the specific settings don't:

<img width="432" height="698" alt="Image" src="https://github.com/user-attachments/assets/66de8075-0c51-4ddd-87a0-cf59625d51f3" />

Issue #666, opened 2024-11-04
Improve error handling for data filtering


ValeriaMaltseva:
**Description:**
Currently, when filtering data, if an error occurs, the error message is displayed but isn't linked to a specific field.

**Task requirements:**
Update the validation error response by linking the error message to the specific field. For example: 
`message: { field: 'fieldName', text: 'text' }`

Current error response:
![Image](https://github.com/user-attachments/assets/83dcda44-7943-4ab0-9c6a-d966ebf54d65)

markus-moser:
Linking it to the field is especially important for the PQL (when a wrong syntax is entered). Let's solve this case (so display the error next to the PQL textarea) and then close this issue.

xIrusux:
Linking the open backend issue that relates to this:

https://github.com/pimcore/studio-backend-bundle/issues/1062

xIrusux:
<img width="1027" height="540" alt="Image" src="https://github.com/user-attachments/assets/57b2f30e-ad69-4540-8a72-b20501d4be0a" />   

<img width="1027" height="540" alt="Image" src="https://github.com/user-attachments/assets/fb899213-9983-4736-847f-0af854727b1c" />  

Tested :). This looks fine now after the backend changes were completed. @heigpa

PR #4088, opened 2026-09-09
[Form Kit] Use spacing tokens and give TabPanel the 8px item gap

idaiv:
Refs #3371

## Scope, and what this does *not* do

#3371 asks for a consistent **8px** vertical gap rule in the form kit. This PR does two things toward that — moves spacing onto tokens, and fixes the one place where the 8px gap was silently not applied at all.

It deliberately **does not** change the value that governs most forms. [`studio-default-light.ts`](../blob/2026.x/assets/js/src/core/modules/app/theme/dynamic-types/definitions/studio-default-light/dynamic-type-theme-studio-default-light.ts#L22-L25) sets `Form.itemMarginBottom: 12`, and **227 files use plain `Form.Item` against 69 using `FormKit`** — so the majority regime in the app is still 12px, not 8px. Changing it shifts every non-FormKit form by 4px and needs a visual pass, so it is left for a follow-up. **The issue should stay open after this merges.**

## Tokens instead of literals

| Was | Now | Pixels |
|---|---|---|
| `padding: '8px 16px'` ×7 (classification-store-config) | `` `${token.paddingXS}px ${token.padding}px` `` | unchanged |
| `GAP_SIZE = 12` (Region) | `token.sizeSM` | unchanged |
| `itemMarginBottom: 8` (language comparison) | `token.marginXS` | unchanged |
| `marginTop: 8` / `fontSize: '12px'` / `color: '#666'` ×2 (appearance color panel) | `marginXS` / `fontSizeSM` / `colorTextSecondary` | unchanged |
| `margin-bottom: 10px` (reports group divider) | `token.marginXS` | **10 → 8px** |
| `margin-top: 50px` ×2 (thumbnail empty states) | `token.marginXXL` | **50 → 48px** |

Everything except the last two rows is pixel-identical by construction. `Region` stays at 12px on purpose — it is a region/column gap feeding a `calc()` width correction, a different axis from field rhythm; tokenising it means the gap and the correction can no longer drift apart.

## TabPanel now supplies the item gap

`FormKit` sets `itemMarginBottom: 0` and expects the gap from `ItemSpacer`, but only `Panel` supplied one. Fields placed straight into a tab therefore rendered **flush** — live example in the class editor: the Text layout type's Configuration tab stacks `renderingClass`, `renderingData` and the WYSIWYG with no gap ([field-definition-text-form-fields.tsx](../blob/2026.x/assets/js/src/core/modules/field-definitions/dynamic-types/types/layout/text/field-definition-text-form-fields.tsx#L84-L112)).

`Tabpanel` now wraps its content in `ItemSpacer`, mirroring `Panel`. Fixed at the component rather than per-caller because **gap only applies between siblings**, so every existing consumer is unaffected:

- `object-tabpanel` — one `<ObjectComponent>` per tab → no change
- all existing stories — `<Panel>` per tab → no change
- `gdpr-data-extractor` — its own local `Tabpanel` (takes `data`/`executeSearch`) → untouched

## Storybook

The TabPanel story now states the contract (8px via `ItemSpacer`, why it is needed given `itemMarginBottom: 0`, and that Panel-wrapping does not double up), plus a new **`FieldsWithoutPanel`** story rendering the shape that used to break, beside a single-field tab so the no-doubling case is visible too.

## Testing

Lint clean on all 13 files; 40 tests pass in `components/form`, `components/tabpanel`, `components/panel`; `npm run build-app` succeeds.

~~**Not verified visually:** `ItemSpacer` introduces a `Flex vertical` wrapper inside every tab, and `object-tabpanel` feeds `ObjectComponent` through it in the data-object editor.~~ **Resolved in bcb599a4f** — the spacer moved to a FormKit-specific `TabPanel`, so `object-tabpanel` keeps the unwrapped shared component and this risk no longer exists.

> Local note: the full-project `npm run lint` and `tsc --noEmit` are OOM-killed in my dev container (6 GiB cgroup, exit 137). Per-file lint is clean and the build passes — CI is the gate for those two.


assets/js/src/core/components/tabpanel/tabpanel.tsx
```diff
@@ -77,8 +78,13 @@ export const Tabpanel = ({
...
       forceRender: true,
       closable: item.closable,
+      // ItemSpacer mirrors Panel: FormKit zeroes antd's itemMarginBottom, so without a spacer
+      // fields placed straight into a tab render flush. A tab holding a single child (a Panel,
+      // an object component) is unaffected — gap only applies between siblings.
       children: <Box padding={ border === true ? 'small' : { x: 'none', y: 'small' } }>
-        {item.children}
+        <ItemSpacer>
+          {item.children}
+        </ItemSpacer>
```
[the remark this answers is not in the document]
idaiv: Good catch — fixed in bcb599a4f, and you were right on both counts.

I had checked the GDPR extractor and wrongly concluded it had its own `Tabpanel`; I'd looked at the wrapper's props (`data`, `executeSearch`) without following the import, which is in fact `Tabpanel as BaseTabPanel` from the shared component. Combined with the SDK re-export at `sdk/components/index.ts:208`, the change would have reached consumers outside FormKit — and worse than an extra wrapper: they keep antd's `itemMarginBottom`, so it would have stacked 12px + 8px instead of establishing 8px.

What changed:

- `components/tabpanel/tabpanel.tsx` is reverted and now byte-identical to its pre-branch state (`git diff ccd2ccd5d^` on that path is empty).
- New `components/form/layouts/tabpanel/form-kit-tabpanel.tsx` maps each tab's `children` through `ItemSpacer` and delegates to `Tabpanel`; `FormKit.TabPanel` points at it instead of being a bare alias. Opt-in, so only forms already running under FormKit's `itemMarginBottom: 0` get the gap.
- The Text layout type was importing the raw `Tabpanel`, so it now imports the FormKit one — otherwise the fix wouldn't have reached the flush Configuration tab this branch set out to fix.

Resulting split: raw shared component for `object-tabpanel`, the GDPR wrapper and the FormKit wrapper itself; spaced variant for the Text layout type and the stories.

One extra benefit of doing it your way: `object-tabpanel` keeps the unwrapped component, so the flex wrapper can no longer affect height/overflow propagation for `ObjectComponent` in the data-object editor — a risk I'd flagged as unverified in the PR description and can now drop.

assets/js/src/core/modules/appearance-branding/components/appearance-form/components/color-panel/color-panel.tsx
```diff
@@ -61,7 +63,7 @@ export const ColorPanel = (): React.JSX.Element => {
               showText
             />
           </Form.Item>
-          <div style={ { marginTop: 8, fontSize: '12px', color: '#666' } }>
+          <div style={ { marginTop: token.marginXS, fontSize: token.fontSizeSM, color: token.colorTextSecondary } }>
```
sholzer:
```diff
-  <div style={ { marginTop: 8, fontSize: '12px', color: '#666' } }>
+  <div style={ { marginTop: token.marginXS, fontSize: token.fontSizeSM, color: token.colorTextSecondary } }>
```

`fontSizeSM` isn't the 12px this replaces. Our theme sets `token.fontSize: 12` (`dynamic-type-theme-studio-default-light.ts:71`), and antd derives `fontSizeSM` one step *below* the configured base — so in Studio:

```
token.fontSize    12
token.fontSizeSM  10
```

Both description lines (66 and 87) end up at 10px instead of 12px. `token.fontSize` is the 12 you want here.

`marginXS` → 8 and `colorTextSecondary` on the same two lines are fine.

sholzer (changes requested):
Traced the TabPanel change end to end — the FormKit-specific split holds up, and I verified all three non-FormKit consumers (object-tabpanel, the object layout type, the GDPR extractor) are genuinely untouched. One token substitution isn't pixel-identical though.

assets/js/src/core/modules/image-thumbnails/image-thumbnails-container.styles.ts
```diff
@@ -10,12 +10,12 @@
 
 import { createStyles } from '@Pimcore/modules/ant-design/styles/create-styles'
 
-export const useStyles = createStyles(({ css }) => {
+export const useStyles = createStyles(({ token, css }) => {
   return {
     emptyState: css`
       text-align: center;
       color: #999;
```
sholzer, line 17: Not needed for this PR, just noticed while reading: this rule uses a fixed `#999`, while the same rule in `video-thumbnails-container.styles.ts` uses `token.colorTextDisabled`. Now that the margin is tokenised in both, the color is the only line left that differs between the two files.

assets/js/src/core/modules/appearance-branding/components/appearance-form/components/color-panel/color-panel.tsx
idaiv: You're right, and thank you — fixed in 101d0e028, both lines now use `token.fontSize`.

I verified it independently rather than taking it on trust, by resolving the tokens against the antd in this repo:

```
with fontSize:12 seed  ->  fontSize: 12   fontSizeSM: 10
antd default           ->  fontSize: 14   fontSizeSM: 12
```

So exactly as you said: 12 → 10 on both description lines.

The mistake is worth naming, because it undercuts a claim I made in the PR description. When I said these substitutions were "pixel-identical by construction", I had checked that `sizeUnit`, `sizeStep`, `paddingXS` and `sizeXS` were not overridden — but never `fontSize`. I reasoned from antd's defaults, where `fontSizeSM` *is* 12, and missed that the theme moves the base out from under it.

I scoped it across the whole branch afterwards: `fontSizeSM` appears only on these two lines. The same resolution confirms the spacing tokens are untouched by that seed (`marginXS: 8`, `paddingXS: 8`, `sizeXS: 8`), so the remaining substitutions really are pixel-identical — as is your read that `marginXS` and `colorTextSecondary` on these lines are fine.

assets/js/src/core/modules/image-thumbnails/image-thumbnails-container.styles.ts
idaiv: Good eye, and you've spotted the seam correctly — but that one is deliberate rather than an oversight, so I've left it as is.

I originally tokenised it to `colorTextDescription` in this branch, which is why the two files briefly matched. Ida reviewed that change and asked for the colour specifically to be reverted, so `#999` is back by request while the margin tokenisation stayed. The video file was never `#999` — it already used `token.colorTextDisabled`, so only its margin changed here.

So the inconsistency you're seeing is real and predates this PR on the image side; I'm just not re-applying a change that was explicitly rejected. Worth noting for whoever picks it up: `#999` on white is roughly 2.8:1, which fails WCAG AA, and #4037 recently lifted `colorTextDisabled` to `rgba(0, 0, 0, 0.55)` (~4.7:1) — so aligning the two files would also close a contrast gap. Happy to do it in a follow-up if you and Ida agree.

sholzer (approved):
LGTM :)

PR #4101, opened 2026-09-15
[Docs] Document entry point providers and the build archive contract

ValeriaMaltseva:
## Summary

Resolves pimcore/service-operations#1345 (PEES-1658). A customer asked what bundles are expected to do to ship a Studio UI frontend that is compatible with Studio and installable via Composer, since neither `WebpackEntryPointProviderInterface` nor `BuildArchiveProviderInterface` were documented, and `BuildArchiveProviderInterface`'s return type (`BuildArchive`) and intended implementation (`BuildArchiveExtractionTrait`) were tagged `@internal` despite the interface itself being public.

`WebpackEntryPointProviderInterface` is already documented (see [Plugins Getting Started](https://docs.pimcore.com/platform/2025.2/Studio_UI/Plugins_Getting_Started/)). `BuildArchiveProviderInterface`, its zip/archive-based distribution mechanism, was not documented anywhere.

Confirmed with @markus-moser: `BuildArchiveProviderInterface` is meant to be public API — community bundles may use it — it was only kept `@internal` during an experimentation phase.

## Changes

- Add `doc/04_Extending/03_Registering_the_Frontend_Build.md`: the provider contract and the two service tags (`pimcore_studio_ui.webpack_entry_point_provider` / `.document_editor_iframe`).
- Add `doc/04_Extending/04_Shipping_the_Frontend_Build_as_an_Archive.md`: the three ways to distribute a compiled Studio frontend (committed build dir, build at deploy time, committed archive) and the step-by-step guide for the archive option (`getBuildGroupId`, `pluginWriteBuildId`, `studio-package-build`, `BuildArchiveProviderInterface`).
- Link the new pages from the Extending index and the Getting Started guide.
- Remove `@internal` from `BuildArchive` and `BuildArchiveExtractionTrait` (the only way to implement `BuildArchiveProviderInterface` without depending on the still-`@internal` `BuildArchiveExtractor` directly), and from the exported `pluginWriteBuildId` rsbuild plugin, which was already reachable externally via the `@pimcore/studio-ui-bundle/rsbuild/plugins` npm export and already used by `copilot-bundle`/`data-importer`. `BuildArchiveExtractor` and `StudioBuildCacheWarmer` stay `@internal`.

Additive only — no behavior change.

## Verification

- PHP syntax lint passes on the touched files.
- All relative links in the touched/added doc pages resolve.
- Docblock lines stay within the 120-char limit.
- Doc content checked against the actual code and against `copilot-bundle`'s real consumption of this mechanism.
- Not run locally: php-cs-fixer and the existing `Build` Codeception unit tests (no PHP tooling installed in this checkout) — left for CI. No new test added since the change is documentation/docblocks only.


ValeriaMaltseva:
Addressed Copilot's review in 1b45e464c:

- **Warmer tag gap**: confirmed real. `StudioBuildCacheWarmer` only wires the main `pimcore_studio_ui.webpack_entry_point_provider` tag (`config/services.yaml:66`); `WebpackEntryPointProviderDocumentEditorIframe` carries only the `.document_editor_iframe` tag and currently avoids the gap only because it shares the same `targetDir` as the main provider — a coincidence, not a guarantee. Documented the restriction rather than changing the DI wiring, since that's a behavior change out of scope for a docs PR.
- **Exception timing**: confirmed. `warmUp()` catches and logs, never rethrows, so `BuildArchiveNotWritableException` only surfaces later at request time, not as a failing `cache:warmup` step. Reworded.
- **Version claim (2025.4.9 vs. `doc/06_Upgrade.md`'s "2025.4.6")**: checked via `git tag --contains` on the introducing commit — `v2025.4.9` is the earliest tag containing it, so 2025.4.9 is correct. `doc/06_Upgrade.md`'s own "Upgrade to 2025.4.6" section header (added in that same commit) looks mislabeled, but that's a pre-existing issue in a file this PR doesn't otherwise touch — flagging it separately rather than fixing it here.

ValeriaMaltseva:
Addressed the second review round in 74cb82131. All three checked out on verification:

- **Archive layout**: confirmed via `unzip -l` on the actually committed archive — it contains `<id>-app/`/`<id>-sdk/`, not a literal `<build-id>/`. Reworded both the doc page and `BuildArchive.php`'s docblock to describe single-target vs. multi-target naming accurately.
- **Missing `pluginWriteBuildId` type declaration**: this is a real, currently-live bug, not a doc nit — I downloaded the actual `@pimcore/studio-ui-bundle@2026.2.10` npm tarball and confirmed `pluginWriteBuildId` is exported at runtime but absent from the published `entrypoints.d.ts`, so the TS import this PR's doc demonstrates would fail to type-check today. Root cause: `assets/dist/types/entrypoints.d.ts` is hand-maintained, not `tsc`-generated (`tsconfig.json`'s `include` only covers `./js/**/*`, never `bundler/`), so it silently drifted out of sync when `pluginWriteBuildId` was added. Fixed by adding the missing declaration; worth a follow-up to make this file generated instead of hand-maintained so it can't drift again.

markus-moser (approved):
As dicussed I have restructered and rewritten the docs page a bit - I decided to extract it into two separate pages. For me it's fine like that now, please check it again, WDYT?

ValeriaMaltseva:
@markus-moser I really like how it's structured now - it is much easier to read and follow. Thanks a lot! I will merge the PR :)

PR #4111, opened 2026-09-17
Place System Settings below Workflow Designer in the main navigation

idaiv:
Resolves #

System Settings appeared **above** Workflow Designer in the System menu, which reads as arbitrary since the two are unrelated.

The cause is a mix of explicit and implicit ordering: `System Settings` pins itself at `order: 600`, while `Workflow Designer` (registered by `pimcore/workflow-designer`) passes no `order` at all and so falls back to the nav registry's default of `1000`:

```ts
// main-nav-registry.ts
items.sort((a, b) => (a.order ?? 1000) - (b.order ?? 1000))
```

Moving System Settings to `1100` puts it after Workflow Designer while staying clear of `About`, which pins itself last at `9999`.

## Additional info
Only the order changes — no markup, permissions or perspective handling is touched.

Fixing it here rather than in `pimcore/workflow-designer` was deliberate: that bundle ships from `vendor/`, so an order added there is not something this repo can rely on, and the implicit `1000` default is what actually decides the position.

Verified by rebuilding Studio locally and checking the menu. Note that `npm run check-types` could not be run in my environment (the TypeScript program exhausts the available heap), but the production `dev-app` build compiled this file successfully.


xIrusux (approved):
LGTM

PR #4112, opened 2026-09-17
Place the ElementTag tooltip to the side, matching the element tree

idaiv:
Resolves pimcore/backend-power-tools-bundle#737

`ElementTag`'s overflow tooltip used antd's default placement, which puts it **above** the tag. In a list of element paths that covers the rows directly above it — exactly the content being compared while scanning the list.

The element tree already places its tooltip on the right (`element-tree-tooltip.tsx`, `placement="right"`). This brings `ElementTag` in line with it.

```diff
-    <Tooltip title={ isOverflow && !pathIsHtml ? path : '' }>
+    <Tooltip
+      placement="right"
+      title={ isOverflow && !pathIsHtml ? path : '' }
+    >
```

## Additional info
**This is intentionally a global change.** `ElementTag` renders its tooltip internally and exposes no placement prop, so the position cannot be set by consumers — the duplicate search list that prompted the issue has no way to override it. Every `ElementTag` tooltip in Studio moves from top to right.

The alternative was adding a `tooltipPlacement` prop and setting it per consumer. I did not take it: it leaves the same misplacement as the default everywhere else, and consuming it from a bundle is gated on an SDK release, so the reporting bundle could not adopt it without a red CI run in the meantime. If reviewers would rather have the prop, say so and I will switch.

Verified by rebuilding Studio locally. `npm run check-types` could not be run in my environment — the TypeScript program exhausts the available heap — but `npm run dev-app` compiled this file successfully and `eslint` is clean on it.

The issue lives in `pimcore/backend-power-tools-bundle`, so the cross-repo reference will not auto-close it on merge.


xIrusux (approved):
LGTM 
:)

PR #4115, opened 2026-09-17
Add the embedding and duplicates configuration icons to the icon library

idaiv:
Resolves pimcore/product-management#1463

Adds the two glyphs for the Backend Power Tools configuration screens to the shared icon set and registers them, so the bundle can reference them by name like any built-in icon.

- `embedding-configuration.inline.svg`
- `duplicates-configuration.inline.svg`

## Additional info
Raised at review on pimcore/backend-power-tools-bundle#773, where the icons were first registered locally in the bundle via `iconLibrary.register(...)`: they belong here instead. That also matches the bundle icons already kept in this set — `alternative-element-tree`, `bookmark`, `bookmark-list-*-folder`.

Registering them centrally has a second benefit: they show up in the icon selector, which a bundle-local registration at plugin init does not reliably give you.

Both use `currentColor` rather than a fixed fill, so they follow the active theme like the rest of the set.

The companion change removing the local registration is on pimcore/backend-power-tools-bundle#773. That bundle references icons **by name at runtime**, so there is no SDK/type dependency between the two — this one just needs to ship first for the glyphs to render.

Verified with `npm run dev-app`; `eslint` clean on the icon library index. (`check-types` does not run in my environment — the TypeScript program exhausts the available heap.)


xIrusux (approved):
Lgtm :)

PR #4117, opened 2026-09-17
fix(alert-modal): treat a null title as no title

idaiv:

Every alert raised through `trackError` without its own title renders with the icon sitting clear of the message it belongs to:

The cause is a `null` that slips past the title guard.

`trackError`'s default path hands the error to `ErrorModalService`, which reports a plain (non-API) error as `{ title: null }` — it has no title to give. `useAlertModal` only guarded against `undefined`:

```ts
title: !isUndefined(title) ? t(title) : t('error')
```

so `null` took the first branch, went through `t()`, and reached antd as a defined title. antd renders its title slot for **any** value that is not `undefined` or `null` (`hasTitle = props.title !== undefined && props.title !== null`), so the dialog gets:

- an empty `.ant-modal-confirm-title` above the message,
- `.ant-modal-confirm-body-has-title`, which aligns the icon to that empty heading rather than to the text,
- the paragraph's `rowGap` between the empty heading and the message.

The icon ends up a row above the first line of text. An error raised *with* a title — every `ApiError` — was unaffected, which is why the standard error alerts look right and only the plain ones do not.

The guard now uses `isNil`, so a `null` title falls back to the same default heading an omitted one already gets. Applied to all four kinds (`info`, `error`, `warn`, `success`) since they share the pattern, and `title` is typed `string | null` to match what callers actually pass.

## Additional info
No call site changes: everything that already passed a title behaves exactly as before, and this is the path `trackError` has always taken.

Covered by `use-alert-modal.test.tsx` — the null-title cases fail against the old guard and pass against the new one.

Found while looking at plain `GeneralError` alerts in the backend power tools bundle, where every one of the 19 call sites hits this path.


xIrusux (approved):
Lgtm :)

